Issue #13, opened 2019-11-04
E2E/Conformance Test Suite


thetechnick:
**User Story**
- Kubernetes ResourceEvent mapping, handling and watching cannot be tested with unittests
- Interaction between multiple controllers cannot be tested with unittests
- Interaction with CRD's and Webhooks cannot be tested with unittests

**Additional**
- The E2E test suite may also be used to test a sponson installation for conformance, similar to the Kubernetes Conformance tests.

**Acceptance criteria**

- The E2E Test Suite should just execute tests against one or more Kubernetes clusters
- Setup of those clusters should be handled in extra tooling
- E2E Test Suite should only require a kubeconfig for the main Kubecarrier cluster and a kubeconfig for each Service Cluster
- E2E Test Suite should be able to be compiled into a CLI subcommand and will be distributed with the CLI
- A E2E Operator should be created and will be used for tests
  - The E2E Operator should not deploy Pods into the cluster, or use other resources than the kube-apiserver
- E2E Tests should be written as independent Suites
  - each suite should test it's requirements/initial state before starting the test run
    (Maybe we should use: https://github.com/stretchr/testify#suite-package for that)
    - preflight checks should be prepared to be executed as part of a CLI command to check if the environment is suitable for the E2E test run
      (Common Interface for Test Suites)

PR #14, opened 2019-11-04
added e2e operator

nmiculinic:
This PR adds the e2e broker according to #13 

Design goals are as follow:

* [x] No pods created, only the kubernetes API objects
* [x] Implements at least 2 different version on each CRD 
* [ ] conversion webhook
* [x] Has override field to showcase provider field overrides in Kubecarrier

Currently the webhook functionality is left for another PR when needed

```release-note
Added test operator purely for end2end test usage
```


nmiculinic:
This is closed in favor of #296 and additional work being done on it in related issues.

PR #105, opened 2019-12-18
WIP -- using server side apply for reconcile

### LINKED ISSUE #80: [Operator] Reduce reconcile API surface
**Is your feature request related to a problem? Please describe.**

`internal/reconcile` API package has large API surface (it has one public function for each supported type)

**Describe the solution you'd like**
Refactor to have only single public API function:

`Reconcile(ctx context.Context, log logr.Logger, c client.Client, desiredObject runtime.Object) error`

everything else is implementation detail.

**Describe alternatives you've considered**
We use reconcile to implement GVK specific update functionality for kubernetes native types....could we reuse kustomize for that as well? In this scenario if the object exists in the cluster it would be the base, and our desired spec would be the patch. This way we reuse kustomize's sane default for patching and are not maintain `reconcile's` complexity

nmiculinic:
TODO

Fixes #80 

```release-note
NONE
```


nmiculinic:
Closed since it's not actively worked on.

PR #276, opened 2020-03-11
Dynamic Elevator Webhook

### LINKED ISSUE #274: Elevator validating/mutating webhooks
Similar to #221, dynamically create webhooks for Elevator components. 
1. Webhooks should be implemented by using dry-run flags to flight against the Catapult webhooks.

jiachengxu:

Fixes #274 

**Documentation**:

```release-note
NONE
```


pkg/internal/resources/catapult/catapult.go
```diff
@@ -233,6 +234,7 @@ func Manifests(c Config) ([]unstructured.Unstructured, error) {
 						},
 					},
 				},
+				SideEffects: &sideEffectsDryRun,
```
nmiculinic: Wait....this is PR regarding elevator, why are you touching catapult code?

nmiculinic: Also which exactly side-effect are we talking about here that you've chosen `NoneOnDryRun` ?

pkg/elevator/internal/webhooks/tenantobj_webhook.go
```diff
@@ -0,0 +1,153 @@
...
+	elevatorutil "github.com/kubermatic/kubecarrier/pkg/elevator/internal/util"
+)
+
+// TenantObjWebhookHandler handles validating of TenantObjs.
+type TenantObjWebhookHandler struct {
+	Log     logr.Logger
+	Scheme  *runtime.Scheme
+	decoder *admission.Decoder
+
+	client.Client
```
nmiculinic: please add some comments/documentation why there are two `client.Client` and when each one should be used. I can definitely see us forgetting the reason a few months in.

pkg/elevator/internal/webhooks/tenantobj_webhook.go
```diff
@@ -0,0 +1,153 @@
...
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/apimachinery/pkg/types"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/webhook/admission"
+
+	catalogv1alpha1 "github.com/kubermatic/kubecarrier/pkg/apis/catalog/v1alpha1"
+	elevatorutil "github.com/kubermatic/kubecarrier/pkg/elevator/internal/util"
+)
+
+// TenantObjWebhookHandler handles validating of TenantObjs.
```
nmiculinic:
```suggestion
// TenantObjWebhookHandler handles TenantObjs validation.
```

Avoid the passive voice as much as possible.

pkg/elevator/internal/webhooks/tenantobj_webhook.go
```diff
@@ -0,0 +1,153 @@
...
+	NamespacedClient client.Client
+
+	TenantGVK, ProviderGVK schema.GroupVersionKind
+
+	DerivedCRName, ProviderNamespace string
+}
+
+var _ admission.Handler = (*TenantObjWebhookHandler)(nil)
+
+// Handle is the function to handle validating requests of ManagementClusterObjs.
```
nmiculinic:
```suggestion
// Handle is the function to handle ManagementClusterObjs validating requests.
```

nmiculinic: Also before you're talking about TenantObjects, and now you're talking about ManagementClusterObjs? 

Furthermore, provider/tenant concepts should be orthogonal to the elevator. It's true that the majority of the code is written with that assumption in mind, thus for consistency we should honor that practice.

pkg/elevator/internal/webhooks/tenantobj_webhook.go
```diff
@@ -0,0 +1,153 @@
...
+	_, otherFields := elevatorutil.SplitStatusFields(exposeConfig.Fields)
+
+	providerObj := &unstructured.Unstructured{}
+	providerObj.SetGroupVersionKind(r.ProviderGVK)
+	providerObj.SetName(obj.GetName())
+	providerObj.SetNamespace(obj.GetNamespace())
+
+	// Check if the ProviderObj has already been created, if it is created, then regards this request
+	// as a UPDATE, if it is not crated, regards this request as a CREATE.
+	err := r.Get(ctx, types.NamespacedName{
```
nmiculinic: shouldn't you get this information upon webhook entry, inside the `req admission.Request`? If the two differ (that is the baseCRD/internal CRD presence is different than derivedCRD/extrenal CRD, we have some serious issues in the kubecarrier system).

All in all, I like this defensive coding approach.

test/integration/derivedcr.go
```diff
@@ -162,6 +162,14 @@ func newDerivedCR(
 			assert.Equal(t, catalogEntry.Status.TenantCRD.Kind, dcr.Status.DerivedCR.Kind)
 			assert.Equal(t, catalogEntry.Status.TenantCRD.Plural, dcr.Status.DerivedCR.Plural)
 		}
+
+		// Check the Elevator dynamic webhook service is deployed.
+		webhookService := &corev1.Service{}
```
nmiculinic: This is a bit too specific for this test. Instead, you should test whether the webhook works properly by issuing a dry-run call/trying to create derivedCRD instance.

test/integration/servicecluster.go
```diff
@@ -194,6 +194,13 @@ func newServiceClusterSuite(
 
 		t.Log("CatalogEntry & CustomResourceDiscoverySet exists")
 
+		// Check the Catapult dynamic webhook service is deployed.
+		webhookService := &corev1.Service{}
```
nmiculinic: This catalog update is orthogonal to the elevator PR it's currenly in. Please move it to separate PR and have it quick lgtmed

pkg/internal/resources/catapult/catapult.go
jiachengxu: This is because the `SideEffects` is `Unknown` by default, and requests with the dryRun attribute will be auto-rejected by the webhook, since we need to make dry-run reqests in the elevator webhook to fight against with the catapult webhook, so we need to specify the `SideEffect` for it.
https://github.com/kubernetes/kubernetes/blob/1c974109b672ade9aae86f2daf1d50c89c36949f/staging/src/k8s.io/api/admissionregistration/v1beta1/types.go#L282

pkg/elevator/internal/webhooks/tenantobj_webhook.go
jiachengxu: Sorry the comment is copied from the `Catapult` webhook, it should be `TenantObjects`

test/integration/derivedcr.go
jiachengxu: yeah, we also need to do the same thing for the catapult PR, since it will also involve e2e operator, so I just want to have a really specific one to check if the webhook service is created, and we can add e2e tests for that later.

pkg/elevator/internal/webhooks/tenantobj_webhook.go
jiachengxu: Using `req.admissinon.Request` was the first attempt in the dynamic PR in the Catapult, and finally, we decided not to use that because it just doesn't work with the `dry-run` requests, I will keep you updated by an example in the following and you will see the problem:

If we use the following for our elevator webhook:
```
switch req.Operation {
  case adminv1beta1.Create:
...
    if err := r.Create(ctx, providerObj, client.DryRunAll); err != nil {
      return admission.Errored(http.StatusInternalServerError, err)
    }
  case adminv1beta1.Update:
...
    if err := r.Update(ctx, providerObj, client.DryRunAll); err != nil {
      return admission.Errored(http.StatusInternalServerError, err)
    }
}
```
The go through the request flow when the tenant tries to create the TenantCRD object:
1. Tenant sends a `CREATE` request,
2. The above webhook regards this as a `CREATE` request, the `DryRun` request can pass (NO problem with this step).
3. TenantObj controller of `Elevator` tries to update the finalizer for the Tenant object, and send a `UPDATE` request. 
4 Then the problem happens: the above webhook regards this as an `UPDATE` request, and the `DryRun` request will fail with `IsNotFound` error, since the `ProviderObject` has not been created.

The similar problem also happens when the `TenantObj` is removed, and update the finalizer later. Also, they are also some other corner cases that can not be handled by this above approach.

There is a walkaround that we can remove the `DryRun` flag and do an actual `Create`/`Update` call, in the webhook, but I just feel like creating objects and introducing some side effects is not the right way to go.

That's why we decided to not use the `req.Operation` but to check if the `ProviderObj` is created or not.

Also, if you think about our approach, it also makes sense, i.e., if the `ProviderObj` is not there, of course it is a `CREATE` request, and it also works fine.

test/integration/servicecluster.go
jiachengxu: moved to #292

pkg/elevator/internal/webhooks/tenantobj_webhook.go
nmiculinic: I can follow steps 1) and 2) for the derivedCR CRD's instance. Step3 is a bit fuzzy to me. What does updating the tenant object's finalizer has anything to do with baseCRD's/derived CR CRD's webhooks? by TenantObj do you mean derivedCR CRD's instance?

nmiculinic: ohhh.... I confused tenant CRD from the `catalog.kubecarrier.io` API group; let me think for a bit more

nmiculinic: Oh, now I see it.

1). Tenant create's tenant object. This send CREATE request on the API server
2) webhook executes with the create, it triggers CREATE with dry-run on the baseCRD (or what you'd call in the code providerObj)
3) webhook successfully executes
4). tenantObj is created.
5). now the derviedCR CRD instance object exists, while the base CRD object doesn't break the invariant on which I assumed my comment. 

Makes sense, please add this explanation in the code documentation, because it's not obvious from the implementation why we chose this approach, and why somebody should not refactor it to use `req.admissinon.Request` in 6 months. 

Great work! Thank you for clearing it up for me.

test/integration/derivedcr.go
nmiculinic: ok, will deal with it later

pkg/elevator/internal/webhooks/tenantobj_webhook.go
nmiculinic: This handles what but is missing the why. They are both operating on the same management cluster, implementing the same interface. The biggest difference is in their caching policy; `client.Client` has global cache, but NamespacedClient has namespace only cache...right?

PR #291, opened 2020-03-13
[WIP] Add multi-region concept

thetechnick:

Fixes #

**Documentation**:

```release-note
NONE
```

### DIFF
--- a/docs/concepts/multi-region.md
+++ b/docs/concepts/multi-region.md
@@ -136,6 +136,7 @@ Features:
   - common list and watch interface across types
 - RBAC should be offloaded to Kubernetes (if extra RBAC is needed)
 - Must retain user-context via e.g. impersonation
+- aggregates catalog information from all `Peers`
 
 ## Tasks
 
--- a/docs/concepts/multi-region.md
+++ b/docs/concepts/multi-region.md
@@ -18,14 +18,12 @@ To drive integration with other tools, KubeCarrier needs to expose an API that c
 - MUST expose watch APIs for caching and wait routines
 - clients MUST be made available for common languages
 - MUST be versioned
-- MUST use user impersonation to retain user context
+- MUST use user impersonation to retain user context for RBAC
+- MUST offer OIDC for user login
 - MUST allow for Service/Robot Accounts
+- MUST offer a REST Interface
 - SHOULD integrate with Kubernetes facilities where possible
   - RBAC, Audit-logging, metric reporting, etc
-
-### TBD
-
-- Compatibility with Web/HTTP Proxies (Protocol Support)
 
 ## Multi Region
 
--- a/docs/concepts/multi-region.md
+++ b/docs/concepts/multi-region.md
@@ -1,0 +1,143 @@
+# KubeCarrier - Multi Region API
+
+## Concepts
+
+### Regions
+
+KubeCarrier treats every `ServiceCluster` as a Region. Each region can have differences in object schemas and are exposed to Tenants. They are similar in scope to regions in AWS or GCP.
+
+### RegionClaims
+
+`RegionClaim` objects claim a region for a KubeCarrier installation and prevent the same `Region` being registered in multiple KubeCarrier `Peers`.
+
+`RegionClaim` objects are Namespaced and belong to Providers.
+
+`ServiceCluster` objects are stuck in a `Pending` state until their `RegionClaim` could is bound and thus acknowledged by the central `KubeCarrier` installation.
+
+```yaml
+apiVersion: kubecarrier.io/v1alpha1
+kind: RegionClaim
+metadata:
+  name: eu-west-1
+  namespace: loodse
+spec: {}
+status:
+  phase: Bound
+  observedGeneration: 0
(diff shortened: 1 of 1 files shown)


nmiculinic:
I've read the proposal, and while I think it's a step in the right direction, I'd prefer it being more hashed out.

(1): What is the reason we want a multi-region concept? That is what issue we are solving and what is the motivation? That should be the start of this document. 

The way I see it the reason/motivation is the following:
* HA/resiliency --> if we have truly global KubeCarrier installation we want cross-region resiliency
* latency --> We want sharding KubeCarrier controller per real geographic region 
* legalese   

This is where the Peer concept falls in nicely, each Peer is responsible for a number of `ServiceClusters`  I dislike the `Region` as defined in the proposal, to me a `Region` would be a set of `ServiceCluster` (not necessarily from the same provider) being managed by a single Peer. Maybe even renaming `Peer` to `RegionController`. 

(2) What are the design requirements for peer/service cluster/account discovery? Are we talking about eventual consistency for CRUD operations, sequential consistency or full-on linearizability? 

(3) If I understood correctly this design assumes the core, alpha & omega master cluster somewhere server as management configuration source of truth for RBAC/Accounts/Peers/etc. Who handles the assignment between `Peers` and `ServiceCluster`s?

docs/concepts/multi-region.md
```diff
@@ -0,0 +1,143 @@
...
+
+The KubeCarrier API will be the main integration point for external tooling to interact with KubeCarrier.
+This contains the CLI, Client-Side Operators, Service Catalog Adapters and UIs.
+
+As the central interface to KubeCarrier, it should hide or abstract complexity as much as possible, making it easy to create integrations on top of it.
+
+Features:
+- routes/aggregates requests to/from `Peers`
+- aggregates the same type across multiple `Regions`/`ServiceClusters`
+  - e.g. exposes `CouchDB.eu-west-1.loodse` and `CouchDB.eu-east-1.loodse` as `CouchDB.loodse`
```
nmiculinic: What about slight version drift issue? That is CRD versions could be the same GVK, but having a slightly different CRD definitions across clusters (e.g. adding single backward compatible field)? IIRC that's the reason we weren't aggregating `ServiceCluster` CRDs into one on single region kubecarrier installation.

docs/concepts/multi-region.md
```diff
@@ -1,4 +1,45 @@
...
+To drive integration with other tools, KubeCarrier needs to expose an API that can be used to drive CLIs, Client-Side Operators, Service Catalog Adapters, UIs and more.
+
+### Goals
+
+- Allows external tools to interact with and extend KubeCarrier
+- Common interface for tooling
+
+### Requirements
+
+- MUST not require to expose the kube-apiserver on public networks
```
nmiculinic: What is the rationale behind this requirement? Will our possible API server be exposed on public networks?

docs/concepts/multi-region.md
```diff
@@ -1,4 +1,45 @@
...
+## Multi Region
+
+A main goal of KubeCarrier is the central management of applications/services across Kubernetes Clusters.
+These Kubernetes Clusters may run on prem, in the cloud, on the edge and across the globe.
+
+To enable KubeCarrier to work in globally distributed and network separated environments and to deal with growth of the system. The KubeCarrier controllers may be distributed across multiple Management clusters.
+
+### Goals
+
+- Enable management clusters (and their Service Clusters) behind firewalls
```
nmiculinic: could you share more details about how the firewalls are setup? What exactly connection is there between API server and management clusters behind firewalls?

docs/concepts/multi-region.md
thetechnick: It seems to be common consensus to not expose the kube-apiserver on the public network, due to security concerns.
Hard to find something in K8s docs, but as reference:

> As the Kubernetes API server isn’t (and shouldn’t be) accessible from the public network, some workloads must use a separate system to authenticate. An example of this is authenticating across clusters, such as from within a cluster to somewhere else.

https://sysdig.com/blog/whats-new-kubernetes-1-18/
#1393 Provide OIDC discovery for service account token issuer

nmiculinic: Can we reuse superCluster kube-apiserver as our root entry point and assume the users shall enforce firewalling kube-apiserver behind VPN

docs/concepts/multi-region.md
nmiculinic: That is the firewalls are done manually, not in an automated fashion with networkPolicy/calico/some other solution.

docs/concepts/multi-region.md
thetechnick: Maybe... although requiring external firewalls and restrictions will make setting up KubeCarrier very complicated

docs/concepts/multi-region.md
thetechnick: I don't think there is currently a solution out there to automate multi-cluster firewalling... maybe Open Service Mesh?
Usually it comes down to normal "legacy" IT processes

PR #292, opened 2020-03-16
Add a test to check if the catapult webhook service is deployed

jiachengxu:

Fixes #

**Documentation**:

```release-note
NONE
```


test/integration/servicecluster.go
```diff
@@ -194,6 +194,13 @@ func newServiceClusterSuite(
 
 		t.Log("CatalogEntry & CustomResourceDiscoverySet exists")
 
+		// Check the Catapult dynamic webhook service is deployed.
+		webhookService := &corev1.Service{}
```
nmiculinic: Can you also check that the service is alive; that is there are valid endpoints for it? 

even better try issuing a request to this service and see if its working.

jiachengxu: Let's put the e2e operator higher in our todo list and check the behave the webhook service once we have it. For now, let's keep it simple and just check the presence of it, is it ok for you?

PR #293, opened 2020-03-16
WIP --> multiregion proposal mark2

nmiculinic:


aborilov:
@nmiculinic 
How aggregating of resources will work?

docs/multiregion.md
```diff
@@ -0,0 +1,98 @@
...
+## Proposed solution -- introducing region concept
+
+To solve the above mentioned issues, we'll introduce managed kubecarrier federation. The basic concepts would be:
+
+* ServiceCluster --> stays the same as before; its akin to availability zone.
+* Region --> a group of multiple providers & service clusters regionally grouped with reasonable low-latency network connection is a region. The region has a single KubeCarrier management cluster
+
+This would provide the logical, geographical, and failure domain grouping to acchive the initial motivational goals. How to coordinate and configure the global KubeCarrier solution from single place (or to the end user appearing as a single place) is still an open problem.
+
+Between regions the following things should be synced, and should be globally unique:
```
jiachengxu: If I understood correctly, you mentioned that every region has a single KubeCarrier management cluster, and it manages a set of service clusters, why does the `ServiceCluster` should also be globally unique?

docs/multiregion.md
```diff
@@ -0,0 +1,98 @@
...
+* https://github.com/kubermatic/kubecarrier/pull/291/files
+
+This solutions revolves around having a single superCluster managing multiple regions with, and each region being managed by region-local installation.
+
+Pros:
+* the user authentication happens at the superCluster level, whilst their actions is propagated to region-local management cluster via the `Impersonate` headers
+* installation is handled via the superCluster operator
+
+Cons:
+* the latency issue still remains. (e.g. request for creating new service instance would propagate from user -- superClusterManager -- region -- region local serviceCluster)
```
jiachengxu: Why the latency issue is still remaining? The creation of the new service instances should not go through the super Cluster, since the accounts are propagated to all KubeCarrier installations.

docs/multiregion.md
```diff
@@ -0,0 +1,98 @@
...
+* the superCluster becomes Single point of Failure
```
jiachengxu: Can we handle that by increasing the replicas of the supercluster installations?

docs/multiregion.md
```diff
@@ -0,0 +1,98 @@
...
+
+### build truly decentralized KubeCarrier installation with helper API/library for cross-region management
+
+The way cross-region requires consists of two operation types:
+* routing
+* fan-out
+
+Also there are two ways for keeping configuration in sync:
+* static configuration
+* central configuration plane (akin to superCluster in the original @nico proposal)
```
jiachengxu: I am a bit lost here, does this mean that we still need to have a central kubecarrier installation? If yes, how this is truly decentralized?

docs/multiregion.md
```diff
@@ -0,0 +1,98 @@
...
+
+The routing operations require the client be routed to the right region executing its request. (e.g. create new serviceCluster, tenant consuming new service, etc.). The typical workflow would work like this:
+
+* client looking up the cache does he have a route in place
```
jiachengxu: How do we keep this cache up-to-date?

docs/multiregion.md
```diff
@@ -0,0 +1,98 @@
...
+    * if not ask the central configuration plane
+    * alternatively fail since the region are hard-coded for the client
+* handle the authentication. The authorization is already handled by the account, and general KubeCarrier installation. The authentication could work in multiple ways:
+    * each kubecarrier region management cluster having the same credentials (e.g. same OIDC settings, or same client CA, or whatever)
+* perform the required operation
+
+The fan-out operations are more complicated, but rarer. They include the following:
```
jiachengxu: I agree with you that adding/removing regions/accounts are rare operations, but why do we introduce extra complexities just for handling the rare operations?

docs/multiregion.md
```diff
@@ -0,0 +1,98 @@
...
+* adding/removing accounts
+* adding/removing serviceCluster (if we want pair account-serviceCluster globally unique)
+
+Their application differ depending on the central configuration store presence. In case the central configuration store is present, add the fan-out operation to the write-ahead log, perform it client side, and report the result. Upon next fan-out operation check the write-ahead log for any possible failure and required clean ups.
+
+In case no central configuration store is present, due to limited operation types that can be done using the known distributed system techniques, it's also a non-issue. For service cluster, for example could work like this:
+
+* append this operation to write-ahead log
+* create ServiceCluster reservation in every cluster
+* convert the target ServiceCluster from reservation to the configuration
```
jiachengxu: Is this like a leader-election?
And what does the `write-ahead` log mean here? Does it present on every KubeCarrier installations, or only on the one with central configuration?

docs/multiregion.md
nmiculinic: I don't know to be honest; that's why I got from @thetechnick proposal that he wants it. 

If I had to pick it usually the region names are globally unique (look at AWS and bigger cloud providers). Also, that way we simplify our CRD names ( <service>.<service cluster>.<provider name> ) and no need adding region in the naming schema.

docs/multiregion.md
nmiculinic: I'm talking about region failure, not a single node failure. It's unlikely, but not an impossible scenario.

docs/multiregion.md
nmiculinic: From what I gather from @thetechnick  proposal all requests are going through the superCluster's Api server which does tha authN/Z, and later on the routing. If not then there's implementation detail how the client discoveres which region to send the request to.

docs/multiregion.md
nmiculinic: You have to somehow handle the region discovery. There are a couple of solutions here, from having it hardcoded in the clients, to having a centralized configuration plane, all the way to DNSesque or BitTorrent peer-discovery decentralized region discovery. Here I'm listing a few options.

docs/multiregion.md
nmiculinic: For routing requests, it's easy --> on cache miss fill in that item.

For fan-out requests, it's more complicated. It's important that all those operations are sequentially consistent, maybe even linearizable. Bad things would happen if regions change concurrently with account addition if there are no eventual consistency guarantees. It can be done, but it's more complicated.

docs/multiregion.md
jiachengxu: but the creation of new service instances should not involve the api-server at all I guess? just like what we have now, having the elevator and catapult running on the KubeCarrier installation for that region, and then tenant can create service isntances I think

docs/multiregion.md
nmiculinic: This is a bit brain-storming. I'm more in favor of having a central configuration plane for regions/accounts that these semi-rare complex operations are handled properly without distributed shenanigans and everything else as distributed as possible.

docs/multiregion.md
nmiculinic: leader-election is different. `write-ahead` log means that you're writing your intentions somewhere, probably in every KubeCarrier installation. It's a bit WIP

PR #311, opened 2020-03-19
Set kubernetes recommanded labels for operator-managed objects

### LINKED ISSUE #278: Use Kubernetes common/recommended labels
Apply Kubernetes common or recommended labels to Operator-managed objects.

| Key | Description |	Example |	Type |
| - | - | - | - |
| app.kubernetes.io/name | The name of the application | mysql | string |
| app.kubernetes.io/instance | A unique name identifying the instance of an application |	wordpress-abcxzy | string |
| app.kubernetes.io/version | The current version of the application (e.g., a semantic version, revision hash, etc.) | 5.7.21 | string |
| app.kubernetes.io/component | The component within the architecture |	database	| string |
| app.kubernetes.io/part-of |	The name of a higher level application this one is part of |	wordpress |	string |
| app.kubernetes.io/managed-by | The tool being used to manage the operation of an application | helm | string |

https://kubernetes.io/docs/concepts/overview/working-with-objects/common-labels/#labels

## Benefits

- Allows to discover dependencies/all objects belonging to a `Ferry`/`Catapult`/`Elevator`/`KubeCarrier` instance
- Allows to easily see the deployed version

## Examples

**Ferry**

```yaml
metadata:
  labels:
    app.kubernetes.io/name: ferry
    app.kubernetes.io/instance: eu-west-1
    app.kubernetes.io/version: v0.1.0
    app.kubernetes.io/managed-by: kubecarrier-operator
```

**Catapult**

```yaml
metadata:
  labels:
    app.kubernetes.io/name: catapult
    app.kubernetes.io/instance: redisinternals.eu-west-1.team-a
    app.kubernetes.io/version: v0.1.0
    app.kubernetes.io/managed-by: kubecarrier-operator
```

jiachengxu:
This PR sets kubernetes recommanded labels for operator-managed objects.
Fixes #278 

**Documentation**:

```release-note
NONE
```


pkg/internal/resources/util/labels.go
```diff
@@ -0,0 +1,27 @@
...
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package util
+
+const (
+	// Kubernetes recommended labels labels: https://kubernetes.io/docs/concepts/overview/working-with-objects/common-labels/#labels
```
nmiculinic: Let's not bloat `util` package even more, please place these labels somewhere else, maybe even have `labels` package which shall store all these string label constants?

pkg/operator/internal/controllers/kubecarrier.go
```diff
@@ -73,6 +73,7 @@ func (c *KubeCarrierStrategy) GetManifests(ctx context.Context, component Compon
 	}
 	return manager.Manifests(
 		manager.Config{
+			Name:      kubeCarrier.Name,
```
nmiculinic: is this some bugfix? If so why wasn't it caught in the e2e tests?

jiachengxu: No this is not, the new `Name` field is added to the `Config` of `KubeCarrier` since the `"app.kubernetes.io/instance"` needs the name

pkg/internal/resources/util/labels.go
nmiculinic: Overall great PR, just address this nit and we're good to go. 

(for acceptance just move all labels somewhere not in `util`, other label refactoring is another isssue/PR)

jiachengxu: I prefer to keep the label as close as possible to the usage of it, this util folder is created in this PR. I don't really see the benefits to keep all labels among different controllers/packages in a cental place.

nmiculinic: `util` is a generic, catch-all name; see https://blog.golang.org/package-names

At least for now move the resource-specific labels to the resource package, e.g. `/pkg/internal/resources/constants` or something like that.

Let's keep the `util` package as lean as possible if we can help it.

PR #313, opened 2020-03-20
CLI: implement the preflight checks

### LINKED ISSUE #53: [Operator/CLI] Prerequisite checks
Prerequisite checks for the KubeCarrier installation.

**Acceptance Criteria**
- checks:
  - Kubernetes v1.16+
  - cert-manager installed (CRD's are installed)
- checks should be checked in init-container of operator
- should be put into shared package, to be also used  by the operator

Hint: This issue should be implemented in mutiple PRs (CLI/Operator)

jiachengxu:
This PR implements a `kubecarrier preflight` command, and also enable the preflight checks in the `kubecarier setup`.
If you do `kubectl kubecarrer preflight` or `kubectl kubecarrier setup`, you can see the following output:
```
0.01s ✔  [preflight check] KubernetesVersion
0.04s ✔  [preflight check] CertManager
```
Currently, there are 2 preflight checks are implemented, one for checking the Kubernetes versions, and one for cert-manager.

Related #53

**Documentation**:

```release-note
NONE
```


nmiculinic:
Two things:
* [ ] preflight checks should be run before `kubectl kubecarrier setup`, that is as part of its functioning. Having a dedicated command for it...maybe. 
* [ ] This PR doesn't completely fix that issue, just implementing the basic minimum in our CLI

pkg/cli/internal/cmd/preflight/checkers.go
```diff
@@ -0,0 +1,167 @@
...
+	utilruntime.Must(corev1.AddToScheme(scheme))
+}
+
+const (
+	firstSupportedKubernetesVersion = "v1.16.0"
+
+	certManagerNamespaceName            = "cert-manager"
+	certManagerCAInjectorDeploymentName = "cert-manager-cainjector"
+	certManagerWebhookDeploymentName    = "cert-manager-webhook"
+	certManagerDeploymentName           = "cert-manager"
```
nmiculinic: We are not going to hard-code how the cert-manager should be deployed in the cluster. That is the responsibility of the system administration setting up the cluster and maintaining it. 

What this check should do is the following --> verify the cert-manager CRDs are present & installed, and at the right version. That's it; everything else is a too strong coupling between how the cert-manager is deployed and kubecarrier.

pkg/cli/internal/cmd/preflight/checkers.go
```diff
@@ -0,0 +1,167 @@
...
+	return nil
+}
+
+// kubernetesVersionChecker checks if the Kubernetes version of the cluster meets the requirement to deploy KubeCarrier.
+type kubernetesVersionChecker struct {
+	firstSupportedVersion string
+	kubernetesVersion     string
+}
+
+func (c *kubernetesVersionChecker) check() error {
```
nmiculinic: how about putting each check in its own file in the same package? one for kubernetes version, one for cert-manager?

pkg/cli/internal/cmd/preflight/checkers.go
```diff
@@ -0,0 +1,167 @@
...
+		}
+	}
+	if errBuffer.Len() > 0 {
+		return fmt.Errorf(errBuffer.String())
+	}
+	return nil
+}
+
+// kubernetesVersionChecker checks if the Kubernetes version of the cluster meets the requirement to deploy KubeCarrier.
+type kubernetesVersionChecker struct {
```
nmiculinic: This shall soonish start being insufficient. We're probably going to have supported some minor version after specific patch versions. (e.g. 1.16.4+ , 1.17.2+ (but not 1.17.0 nor 1.17.1) ) )

pkg/cli/internal/cmd/preflight/checkers.go
jiachengxu: but it can be the case that cert-manager CRDs are installed but the deployments are not ready, which means even if the preflight checks are passed, the `kubecarrier install setup` will still fail.
Also for checking the CRDs, I also need to hard-code the CRD names, right?

jiachengxu: what do you think @thetechnick @aborilov ?

pkg/cli/internal/cmd/preflight/checkers.go
jiachengxu: I don't really know what you mean? 1.16.4 <= 1.17.2? For this case, it works

jiachengxu: Do you mean we need to compare some versions like 1.16.4 and 1.17.2?

jiachengxu: also there is a test about it https://github.com/kubermatic/kubecarrier/pull/313/files#diff-2152533908d9e0e3e13faa9697127d76R64

nmiculinic: > I don't really know what you mean? 1.16.4 <= 1.17.2? For this case, it works

but it doesn't exclude 1.17.0 and 1.17.1 as an unsupported version, because for example, they don't include some bugfix we need for proper functioning.

pkg/cli/internal/cmd/preflight/checkers.go
nmiculinic: CRD names are cluster scoped and more stable. Furthermore, we explicitly depend on that CRD names when creating cert-manager objects in the kubecarrier itself. Thus it's a direct dependency that those CRDs exist.

nmiculinic: > but it can be the case that cert-manager CRDs are installed but the deployments are not ready, which means even if the preflight checks are passed

that is true. But with your approach, you're enforcing exact deployment name & namespace. You could still do this, but mark it only as a warning if they are not there. Something like. 

`[WARN] cert-manager deployment not found in cert-manager namespace. The kubecarrier requires working cert-manager deployment. You're responsible for satisfying this condition.`

pkg/cli/internal/cmd/preflight/checkers.go
jiachengxu: Ok, I see, should we cover this case also @thetechnick ?

pkg/cli/internal/cmd/preflight/checkers.go
jiachengxu: `cert-manger` namespace is the default namespace to install cert-manager from cert-manger official document: https://cert-manager.io/docs/installation/kubernetes/#installing-with-regular-manifests
They also mentioned the following notes:

Note: By default, cert-manager will be installed into the cert-manager namespace. It is possible to run cert-manager in a different namespace, although you will need to make modifications to the deployment manifests.

I am not really sure this will happen a lot, since I don't really know what is the use case for some people to deploy cert-manger in some other namesapce, and introduce extra complexity to edit the manifests.

What do you think @thetechnick ?

pkg/internal/util/clientwatcher.go
```diff
@@ -55,7 +55,7 @@ func WithoutClientWatcher() ClientWatcherOption {
 }
 
 const (
-	defaultTimeout = 30 * time.Second
+	defaultTimeout = 60 * time.Second
```
nmiculinic: please don't increase the defaultTimeout for all in this PR.

jiachengxu: Why? Many tests are flaky due to the short timeout, not only on prow and my local machine, I don't want to type `\retest` all the time.

nmiculinic: 30s is quite a lot of time. Are they flacky on everything or just some specific operation? If so we should only increase that slow operation, not everything.

pkg/cli/internal/cmd/preflight/checkers.go
thetechnick: I also think that checking CRDs is enough.
If we need more detailed error handling, the KubeCarrier Operator should report more status details, like: `CertificatesReady` or similar. That way we can show meaningful error messages to the enduser, without having us to rely on too much hardcoded information.

pkg/cli/internal/cmd/preflight/checkers.go
thetechnick: Let's validate for the lowest supported version first, we can maintain a blacklist of "broken" k8s versions, if we have a case for it. Currently there is no need to add additional complexity here.

PR #315, opened 2020-03-23
CustomResourceDiscovery deletion webhook

### LINKED ISSUE #280: [Bug] CustromResourceDisovery deletion leads to stuck state
**Describe the bug**
Upon deleting `CustromResourceDiscovery` with created CRD instances in the management cluster there are no errors, but the system gets stuck in the deadlock state (without manual intervention).

Here's what's happening. The `CustomResourceDiscovery` deletion triggers the following:

* the `Catapult` is concurrently being deleted since it has `CustromResrouceDiscovery` as its owner (Kubernetes native GC, `metadata.ownerReferences` fields )
* there's catapult's finalizer on the created internal CRD instances
* The operator clears out the catapult deployment, leading all its catapult related finalizer impossible to satisfy (including those on the created CRD instances)
* The CRD deletion process starts, and there's `customresourcecleanup.apiextensions.k8s.io` finalizer waiting for all its instances deleted before the `CustromResourceDiscovery`'s internal CRD is deleted.
* those CRDs cannot be deleted since they depend on catapult deleting base CRDs on the service cluster and then removing its finalizer
* But that isn't possible since the catapult deployment is deleted.
* All in all, a messy place to be all around. 

**To Reproduce**
1. create `CustromResourceDiscovery`
2. create `CustromResourceDiscovery` CRD instance in the management cluster 
3. delete `CustromResourceDiscovery`
4. Observe error

**Possible fixes**

* operator should check that all objects having the specific catapult finalizer are cleared before deleting its deployment (( and optionally having different finalizers per catapult deployment, thus being able differentiating between them ))
* a webhook should prevent the  `CustromResourceDiscovery` deletion if any instances are present at the moment ((probably the smartest idea))

jiachengxu:

Fixes #280

**Documentation**:

```release-note
NONE
```


pkg/manager/internal/webhooks/customresourcediscovery_webhook.go
```diff
@@ -20,25 +20,35 @@ import (
...
 
 var _ admission.Handler = (*CustomResourceDiscoveryWebhookHandler)(nil)
 
-// +kubebuilder:webhook:path=/validate-kubecarrier-io-v1alpha1-customresourcediscovery,mutating=false,failurePolicy=fail,groups=kubecarrier.io,resources=customresourcediscoveries,verbs=update,versions=v1alpha1,name=vcustomresourcediscovery.kubecarrier.io
+// +kubebuilder:webhook:path=/validate-kubecarrier-io-v1alpha1-customresourcediscovery,mutating=false,failurePolicy=fail,groups=kubecarrier.io,resources=customresourcediscoveries,verbs=update;delete,versions=v1alpha1,name=vcustomresourcediscovery.kubecarrier.io
 
-// Handle is the function to handle create/update requests of CustomResourceDiscoveries.
+// Handle is the function to handle update/delete requests of CustomResourceDiscovery objects.
 func (r *CustomResourceDiscoveryWebhookHandler) Handle(ctx context.Context, req admission.Request) admission.Response {
 	obj := &corev1alpha1.CustomResourceDiscovery{}
```
nmiculinic: IIRC delete admission hook doesn't have current obj setup, thus this would error out. Move this to the `UPDATE` section.

test/integration/servicecluster.go
```diff
@@ -148,7 +148,7 @@ func newServiceClusterSuite(
 			},
 		}
 		require.NoError(t, managementClient.Create(ctx, catalogEntrySet))
-		require.NoError(t, testutil.WaitUntilReady(ctx, managementClient, catalogEntrySet))
+		require.NoError(t, testutil.WaitUntilReady(ctx, managementClient, catalogEntrySet), testutil.WithTimeout(time.Minute))
```
nmiculinic: how bout making a dedicated integration test suite for the `CustomResourceDiscovery`? Similar to how `DerivedCustomResource` has its own dedicated suite. Especially since we're going to test more for this integration suite than what should be in scope for the service cluster.

jiachengxu: but if you look into the servicecluser suite closely, it is integration test suite for `CustomResourceDiscovery`,  it is called `servicecluster` because it is doing some cross-cluster tests IMO.

nmiculinic: ok; seems good to me at least for this PR.

PR #323, opened 2020-03-25
WIP catalogEntrySet deletion bug

nmiculinic:

Recently in the flaky e2e test CatalogEntrySet cannot be cleared. That is because the following two things are in a race condition:
* deleting clusterRole allowing manager listing derivedCR resources
* deleting `DerivedCustomResource` which triggers the webhook in the manager, which lists the derivedCR resources for existence. Thus without the former role that's impossible to do. 

This PR moves ClusterRole creation from the elevator/catapult in the manager's controller before the deletion webhook is loaded. 

```release-note
NONE
```


pkg/apis/catalog/v1alpha1/derivedcustomresource_types.go
```diff
@@ -184,6 +184,9 @@ func (s *DerivedCustomResource) IsReady() bool {
 	if s.Generation != s.Status.ObservedGeneration {
 		return false
 	}
+	if !s.DeletionTimestamp.IsZero() {
```
nmiculinic: This fix forbids creating new derived custom resources if the `DerivedCustromResource` CRD instance is under deletion.

nmiculinic:
Closing for now because #336 got merged.

PR #353, opened 2020-03-30
Local ManagementCluster

### LINKED ISSUE #344: Tower - Local ManagementCluster
`Peer` objects are cluster-scoped and track management clusters that are part of this multi-region deployment.

See the Epic for further details.

This task should add support to specify the current/local cluster as a `Peer`.

Acceptance Criteria:
- API group `master.kubecarrier.io` exists
- `ManagementCluster.master.kubecarrier.io` type exists
- the `local` Peer is always reported as ready

Example:

```yaml
apiVersion: master.kubecarrier.io/v1alpha1
kind: ManagementCluster
metadata:
  name: local
spec: {}
status:
  phase: Ready
  conditions: []
  observedGeneration: 0
```

jiachengxu:
This PR adds the implementation of managing the local `ManagementCluster` object of KubeCarrier installation with `master:true`.
For creating a master KubeCarrier installation, just use:
```
kubectl kubecarrier setup --master
```
1. A `KubeCarrier` object with `master:true` will be created.
2. The multi-region microservice (Tower) will be deployed.
3. The `ManagementCluster` with name `local` will be created, and will be reported as `Ready` by controller of Tower.

Fixes #344 

**Documentation**:

```release-note
NONE
```


nmiculinic:
The PR is technically fine, though I'm kinda concerned about the workflow. The way I see it, our federation would work in the following manner:

* having dedicated master cluster overseeing multiple management clusters. Should this cluster have deployed the manager controller? I don't think so....it should only consist of an operator, tower, api-server, and the CRDs. That's the recommended approach
* have small installations/development setup where all components are on the same cluster, that is where the master cluster is also the only management cluster. 

What do you think about this approach? In this PR you assume master cluster is also a local management cluster, which isn't desirable in the enterprise setup.

jiachengxu:
> having dedicated master cluster overseeing multiple management clusters. Should this cluster have deployed the manager controller? I don't think so....it should only consist of an operator, tower, api-server, and the CRDs. That's the recommended approach

What is the drawback that the manager controller is also deployed? As you mentioned above, we also  should support running all components on the same cluster, this just provides the way of doing that.

> In this PR you assume master cluster is also a local management cluster, which isn't desirable in the enterprise setup.
Why is this not desirable in the enterprise setup? Can you please provide an example here?

thetechnick:
Needs to target `feature/master-controller-manager` after #382 is merged.

thetechnick:
/lgtm
/hold
:+1:

PR #363, opened 2020-03-30
Operator Refactor

### LINKED ISSUE #352: [Operator] Hard to extend operator controllers with more specific reconcile logic.
a base_reconciler was introduced in #250, and it works fine for our current components (KubeCarrier, ferry, catapult, elevator)

However, now we are going to add our new component (Tower) and makes the KubeCarrier object cluster-scoped, then the issues are coming out:
# Making KubeCarrier cluster-scoped: #329 
Once the `KubeCarrier` becomes cluster-scoped, there is no need to use our own owner handling, we can use `contrillerutil.SetControllerReference` to handle the owned objects, this lead the first issue in the base reconciler:
```
// ReconcileOwnedObjects - reconcile the objects that owned by obj
func (r *BaseReconciler) reconcileOwnedObjects(ctx context.Context, obj Component, objects []unstructured.Unstructured) (bool, error) {
	var deploymentIsReady bool
	for _, object := range objects {
...
		switch restMapping.Scope.Name() {
		case meta.RESTScopeNameNamespace:
			if err := controllerutil.SetControllerReference(obj, &object, r.Scheme); err != nil {
				return false, err
			}
		case meta.RESTScopeNameRoot:
			owner.SetOwnerReference(obj, &object, r.Scheme)
		default:
			return false, fmt.Errorf("unknown REST scope: %s", restMapping.Scope.Name())
		}
...
	}
	return deploymentIsReady, nil
}
```
In the function above, we only check if the owned object is namespaced or cluster-scoped, and then choose different owner handling methods accordingly. But for the owner object, we didn't check anything. In PR #347, I added a check to check if the owner object is cluster-scoped (KubeCarrier) by checking if the namespace is empty:
```
// ReconcileOwnedObjects - reconcile the objects that owned by obj
func (r *BaseReconciler) reconcileOwnedObjects(ctx context.Context, obj Component, objects []unstructured.Unstructured) (bool, error) {
...
		if obj.GetNamespace() == "" {
			// The owner is cluster-scoped, like KubeCarrier.
			if err := controllerutil.SetControllerReference(obj, &object, r.Scheme); err != nil {
				return false, err
			}
		} else {
			switch restMapping.Scope.Name() {
			case meta.RESTScopeNameNamespace:
				if err := controllerutil.SetControllerReference(obj, &object, r.Scheme); err != nil {
					return false, err
				}
			case meta.RESTScopeNameRoot:
				owner.SetOwnerReference(obj, &object, r.Scheme)
			default:
				return false, fmt.Errorf("unknown REST scope: %s", restMapping.Scope.Name())
			}
		}
...
	}
	return deploymentIsReady, nil
}
```
This works, but not works nicely.

# Start Tower in KubeCarrier controller #325 
The issue is even worse once we now want to introduce the `Tower` for our multi-region support. 
Now we want to have a `master` flag in KubeCarrier object:
```yaml
apiVersion: operator.kubecarrier.io/v1alpha1
kind: KubeCarrier
metadata:
  name: kubecarrier
spec:
  master: true
```
If the `master` is `true`, then the `KubeCarrier` controller will create `Tower` object, and deploy `Tower`. This is even harder to achieve in the operator base reconciler.
It can be solved by adding a checking like, if the owner is `KubeCarrier`, if it is, then starts Tower...
but this doesn't solve the issue from the root at all, and we may have more similar issues in the future when we want to add more specific reconciliation logic in the future.

# Solution
Currently, I didn't see any better solution than reverting the base_conciler back to several controllers. Although there will be some code duplications, but at the least, the code is more extendable.

jiachengxu:
This PR refactors the operator controllers to make it easy to extend.

Fixes #352 

**Documentation**:

```release-note
NONE
```


aborilov:
Looks good and will work, but I think it can be done easier and without changing all other operators. If we make `BaseReconciler` methods to be public, we can create an operator with any logic and reuse existing `BaseReconciler` logic in it.

here is my proposal https://github.com/kubermatic/kubecarrier/pull/367

nmiculinic:
/hold

As I've talked with @jiachengxu  yesterday, I find this refactoring unnecessary and counterproductive. For now, all our use-cases are covered, and the bigger underlying issue is our byzantine owner handling functionality which we should extract in a few clean helper functions. Those helper functionss shall know how to properly set owner annotations/labels/metadata.ownerReference depending on the clusterScope/namespaceScope of the object and the owner object. 

Furthermore we should make a enqueuerForOwner which works with all 3 approaches (owner annotation, label owners, metadata.ownerReferences) and just use that one as catch all. This solves most of this refactorings usecase and benefits.

jiachengxu:
@nmiculinic, you can always make it work, but it is the way complicated to make it work than you thought, in the following, I will show you what changes should be made to implement the `Starting Tower in KubeCarrier controller`, you can check the code in Local ManagementCluster PR #353

1. You need to convert `Tower` object to `unstructured.Unstructured` in the `GetManifest` method:
https://github.com/kubermatic/kubecarrier/blob/611ba7e85a931eab8a8b7c6ac9682b2229db9763/pkg/operator/internal/controllers/kubecarrier.go#L84-L101

2. You need to implement the reconcile functions in `internal/reconcile` package to reconcile `Tower`:
https://github.com/kubermatic/kubecarrier/blob/611ba7e85a931eab8a8b7c6ac9682b2229db9763/pkg/internal/reconcile/operator.go#L33-L71

3. You need to implement the convert method to convert `unstructured.Unstructured` object to `Tower` object:
https://github.com/kubermatic/kubecarrier/blob/611ba7e85a931eab8a8b7c6ac9682b2229db9763/pkg/internal/reconcile/unstructured.go#L332-L344

4. You need to register the reconcile method in the `unstructuredReconcilers`:
https://github.com/kubermatic/kubecarrier/blob/611ba7e85a931eab8a8b7c6ac9682b2229db9763/pkg/internal/reconcile/unstructured.go#L132-L137

Please notice that, if you missing any one of the above steps, it will not work. This is how to extend the operator controllers, you need to implement 4 steps which are quite decentralized, to just start `Tower` in `KubeCarrier`. In the future, if we want to extend with more complicated logic, I don't want to implement the above 4 steps again.

What if we have a dedicated kubecarrier controller?  
Create `Tower` object in it, set ownerreference, and use `owner.ReconcileOwnedObjects` to reconcile, that's about it. Don't need to convert Tower object to `unstructured.Unstructured` object, and implement the 4 places which are mentioned above. Also, all the logic is in the kubecarrier contrller, which is easy to debug.

That's I stick to refactor the operator controller, it just easy to extend, I see no reason to have  a centralized base controller.

WDYT @thetechnick @aborilov

aborilov:
@jiachengxu but you can just add `kubecarrier_controller` with your own `Reconcile` method, as I did in my PR https://github.com/kubermatic/kubecarrier/pull/367, it's a very small change, why do you need to refactor all other controllers?

jiachengxu:
> @jiachengxu but you can just add `kubecarrier_controller` with your own `Reconcile` method, as I did in my PR #367, it's a very small change, why do you need to refactor all other controllers?

Hmmm, but why do we keep the base_reconciler if it is not the base of all our operator controllers?

aborilov:
> > @jiachengxu but you can just add `kubecarrier_controller` with your own `Reconcile` method, as I did in my PR #367, it's a very small change, why do you need to refactor all other controllers?
> 
> Hmmm, but why do we keep the base_reconciler if it is not the base of all our operator controllers?

it's base for 4 of 5 operator controllers and I think you still can use some of its methods for `kubecarrier` controller

jiachengxu:
> I'm all for improving the codebase. Though operator refactors isn't the right place for it, but 1) working with properly go typed runtime.Object and 2,3,4) forgetting about unnecessary duplicate logic and just make a controller who's gonna reconcile all non-metadata non-status fields on update.

But now the case is even worse... The tower has status, and I need to check if the is ready, so I can update the `KubeCarrier` status accordingly....

aborilov:
> > I'm all for improving the codebase. Though operator refactors isn't the right place for it, but 1) working with properly go typed runtime.Object and 2,3,4) forgetting about unnecessary duplicate logic and just make a controller who's gonna reconcile all non-metadata non-status fields on update.
> 
> But now the case is even worse... The tower has status, and I need to check if the is ready, so I can update the `KubeCarrier` status accordingly....
> How can I implement that with current base_reconcler?
you can check it in your `Reconcile` method and then call `BaseReconciler.UpdateStatus`.
if you need some specific logic, just add it in your methods

jiachengxu:
> you can check it in your `Reconcile` method and then call `BaseReconciler.UpdateStatus`.
> if you need some specific logic, just add it in your methods

Are you talking about the refactor in #367 or current codebase?
I didn't see any possibilities to implement based on our current codebase, what I want to achieve is as following, not only set ready or unready condition:
having a dedicated condition type `TowerReady`, and set it to `True` if the `Tower` is ready,
having a dedicated condition type `KubeCarrierDeploymentReady` and set it to `True` if the `KubeCarrier` Deployment is ready.

If both of the above conditions are true, then set ready condition for KubeCarrier.

You can find the similar logic in https://github.com/kubermatic/kubecarrier/blob/master/pkg/manager/internal/controllers/customresourcediscovery_controller.go#L285-L327 and https://github.com/kubermatic/kubecarrier/blob/master/pkg/manager/internal/controllers/derivedcustomresource_controller.go#L396-L437
In this way, we can provide more useful info to the kubecarrier admin, and also it is easier to debug

nmiculinic:
Why do you need deconstructing everything when there's an adaptive solution. Currently, we have the following two interfaces in the base reconciler:

```go
type Component interface {
	runtime.Object
	metav1.Object
	SetTerminatingCondition() bool
	SetUnReadyCondition() bool
	SetReadyCondition() bool
}

type ControllerStrategy interface {
	// GetObj - return origin controller object
	GetObj() Component
	GetManifests(context.Context, Component) ([]unstructured.Unstructured, error)
	GetDeletionObjectTypes() []runtime.Object
	AddWatches(builder *builder.Builder, scheme *runtime.Scheme) *builder.Builder
}
```

To satisfy your new requirements we could do the following. Change the interface to the following:

```go
type Component interface {
	runtime.Object
	metav1.Object
        UpdateStatus(ctx context.Context, cl client.Client) error
}

type ControllerStrategy interface {
	// GetObj - return origin controller object
	GetObj() Component
	GetManifests(context.Context, Component) ([]unstructured.Unstructured, error)
	GetDeletionObjectTypes() []runtime.Object
	AddWatches(builder *builder.Builder, scheme *runtime.Scheme) *builder.Builder
}
```

for example, and the `UpdateStatus` could check all necessary conditions (in the usual case whether the specific deployment is ready, in the case you proposed also check the Tower's status too), and we're still remaining in a nice framework like approach. 

If you'd like status updating more in the `ControllerStrategy` (because the Component entails changing the API's go file and the logic there should be minimal anyway) we can do something like this:

 ```go
type Component interface {
	runtime.Object
	metav1.Object
}

type ControllerStrategy interface {
	// GetObj - return origin controller object
	GetObj() Component
	GetManifests(context.Context, Component) ([]unstructured.Unstructured, error)
	GetDeletionObjectTypes() []runtime.Object
	AddWatches(builder *builder.Builder, scheme *runtime.Scheme) *builder.Builder
        UpdateStatus(context.Context, cl client.Client, Component) error
}
```

without the need to deconstruct what we already achieved. What is great about the framework approach is enforcing the implicit structure we already have in our controller, not by convention during PR reviews, but by the compiler. That way all controllers follow the similarly named function and when looking at them I'm aware where to look for which functionality. 

All our controllers, in general, have the same basic loop encoded in the base reconciler:
* finalizer handling
* deletion handler --> killing everything I've created for this object
* create all objects from some manifests
* update manifest depending on few objects status

and this base reconciler fits it nicely.

jiachengxu:
> Why do you need deconstructing everything when there's an adaptive solution. 

Because the adoption you mentioned above doesn't reduce the complexities I mentioned in https://github.com/kubermatic/kubecarrier/pull/363#issuecomment-606546993, and I don't agree that the main reason is that `reconcile`  and `resource` packages have mistakes or whatever(of course they can be improved, but that is not the reason that operator controller is difficult to extend), at least other controllers work smoothly. I don't like the idea to have a base reconciler, and then in the future, every time you need to introduce extra complexities, or you spend plenty of time to adopt it, it is completely overkilled for adding new features, I like abstracting code, but I don't accept the code which is hard to extend.

jiachengxu:
Also, since the `KubeCarrier` object is now cluster-scoped, there is no need to use finalizer for it, please take a look at the `KubeCarrier` controller, and check how simple it is.

nmiculinic:
I mean...let's see what @thetechnick  thinks about it. The proposed refactoring controllers are mostly c/p code for one to the other with the common structure not being enforced.

thetechnick:
I have looked though both this and https://github.com/kubermatic/kubecarrier/pull/367 / the existing operator code.

I no big fan of the current (master) operator refactoring, because it assumes that all operator controllers will stay the same in the future, which is a dangerous assumption. This way of refactoring is IMHO more suitable to code that is put into a maintenance mode, where only little changes and bug fixes are expected.

I consider especially our Operator-Controllers as one of the weakest controllers that we have, they still need a lot of functionality which will involve experimentation and refactoring on a per-controller basis.

> I mean...let's see what @thetechnick thinks about it. The proposed refactoring controllers are mostly c/p code for one to the other **with the common structure not being enforced.**

Thats especially what I don't want in the project at the moment. Enforcing a common structure across multiple *Classes or controllers makes only sense to me after the components are mostly feature complete and you are not only interested in removing duplicated functionality, but you also want to lower your LoC to ease in reviews.

For our current situation I am favouring this PR.

Let's discuss this again in our daily on Monday and get one of these PR's merged.
@jiachengxu @nmiculinic @aborilov

PR #365, opened 2020-03-30
API server bootstrapping

### LINKED ISSUE #354: [API server] Bootstrap initial API server
**Is your feature request related to a problem? Please describe.**
The API server has to be bootstrapped somehow.

**Describe the solution you'd like**
* [ ] Initial API server CRD. It should belong to `master.kubecarrier.io` API group
* [ ] the operator should reconcile the API server CRD into appropriate deployment, service, and even an ingress resource
* [ ] The server in its initial version should be nothing but bare HTTP/gRPC server with dummy implementations

nmiculinic:

Implement initial bootstrapping for the API server

Fixes #354 

```release-note
Implement initial bootstrapping for the API server
```


config/internal/api-server/default/kustomization.yaml
```diff
@@ -0,0 +1,70 @@
+# Adds namespace to all resources.
+namespace: kubecarrier-system
+
+# Value of this field is prepended to the
+# names of all resources, e.g. a deployment named
+# "wordpress" becomes "alices-wordpress".
+# Note that it should also match with the prefix (text before '-') of the namespace
+# field above.
+namePrefix: kubecarrier-tower-
```
jiachengxu: can you change the tower to api-server?

jiachengxu: And also for other yaml files?

hack/gen-tasks.go
```diff
@@ -142,15 +142,16 @@ func main() {
...
-		"Integration/catalog",
 		"Integration/account",
+		"Integration/api-server",
+		"Integration/catalog",
 		"Integration/cli",
+		"Integration/derivedCR",
+		"Integration/serviceCluster",
 		"Scenarios",
 		"Scenarios/simple",
+		"VerifySuite",
```
jiachengxu, line 154: Why this `VerifySuite` is moved to the last test?

pkg/api-server/apiserver.go
```diff
@@ -0,0 +1,78 @@
...
+	log.Info("booting serving API-server", "port", flags.port)
+	http.HandleFunc("/", func(writer http.ResponseWriter, request *http.Request) {
+		body, err := ioutil.ReadAll(request.Body)
+		if err != nil {
+			writer.WriteHeader(http.StatusInternalServerError)
+			fmt.Fprintln(writer, err.Error())
+			return
+		}
+		log.Info("new request", "body", string(body), "method", request.Method, "host", request.Host, "path", request.URL.Path)
+		fmt.Fprintln(writer, "hello world!")
```
jiachengxu, line 75: please remove this line

pkg/api-server/apiserver.go
```diff
@@ -0,0 +1,78 @@
...
+var (
+	scheme = runtime.NewScheme()
+)
+
+func init() {
+	utilruntime.Must(clientgoscheme.AddToScheme(scheme))
+}
+
+const (
+	componentTower = "Tower"
```
jiachengxu, line 43: I don't think the api-server is named `Tower`, the master controller manager is named tower

pkg/apis/operator/v1alpha1/apiserver_types.go
```diff
@@ -0,0 +1,232 @@
...
+			s.Conditions[i].Message = condition.Message
+
+			return
+		}
+	}
+
+	s.Conditions = append(s.Conditions, condition)
+}
+
+// APIServer manages the deployment of the KubeCarrier master controller manager.
```
jiachengxu, line 149: Please update this comment

pkg/operator/internal/controllers/api-server.go
```diff
@@ -0,0 +1,75 @@
...
+
+	appsv1 "k8s.io/api/apps/v1"
+	corev1 "k8s.io/api/core/v1"
+	rbacv1 "k8s.io/api/rbac/v1"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/runtime"
+	"sigs.k8s.io/controller-runtime/pkg/builder"
+
+	operatorv1alpha1 "github.com/kubermatic/kubecarrier/pkg/apis/operator/v1alpha1"
+	resourcetower "github.com/kubermatic/kubecarrier/pkg/internal/resources/tower"
```
jiachengxu, line 31: This is not working... why you are not using "github.com/kubermatic/kubecarrier/pkg/internal/resources/apiserver"
Can you please add a test that this will never happen in the future?

hack/gen-tasks.go
nmiculinic: alphabetic sorting, as it should have been done in the first place.

pkg/operator/internal/controllers/api-server.go
nmiculinic: I see...some of this stuff is fixed in the gRPC commits...will backport

pkg/api-server/apiserver.go
nmiculinic: No, this is fine and expected.

pkg/operator/internal/controllers/api-server.go
```diff
@@ -0,0 +1,75 @@
...
+
+	appsv1 "k8s.io/api/apps/v1"
+	corev1 "k8s.io/api/core/v1"
+	rbacv1 "k8s.io/api/rbac/v1"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/runtime"
+	"sigs.k8s.io/controller-runtime/pkg/builder"
+
+	operatorv1alpha1 "github.com/kubermatic/kubecarrier/pkg/apis/operator/v1alpha1"
+	apiserver "github.com/kubermatic/kubecarrier/pkg/internal/resources/api-server"
```
jiachengxu, line 31: Can you please add a test for it to avoid this kind of mistake that will never happen in the future? 
I had some mistake in the catalog controller and fixed in this PR: https://github.com/kubermatic/kubecarrier/pull/231,
I think for this case, a test is necessary to detect this issue in the future

pkg/internal/resources/api-server/apiserver.go
```diff
@@ -0,0 +1,76 @@
...
+	"github.com/kubermatic/kubecarrier/pkg/internal/kustomize"
+	"github.com/kubermatic/kubecarrier/pkg/internal/resources/constants"
+	"github.com/kubermatic/kubecarrier/pkg/internal/version"
+)
+
+// Config holds the config information to generate the KubeCarrier master controller manager setup.
+type Config struct {
+	// Namespace is the KubeCarrier master controller manager should be deployed into.
+	Namespace string
+	// Name of this Tower object.
```
jiachengxu, line 35: Please pay more attention to your PR... and remove all the occurrences of `tower`.

nmiculinic: I did...but forgot the resources subpackage.

pkg/operator/internal/controllers/api-server.go
nmiculinic: I don't think it makes sense for this PR...and furthermore as we add more features these mistakes are more obvious.

jiachengxu: No, I don't think the mistakes will be obvious, because actually this wrong import will:
1. miss the CRDs of api-server, which is hard to debug since you can see the CRDs are generated correctly in golden file.
2. override Tower configurations (ClusterRole, ClusterRoleBinding...) which is even harder to debug, because Tower and API server will override the same config all the time. We had the same issue of when the oeprator and manager orerrided the same secret to deploy webhooks, and it was really hard to debug.

pkg/api-server/apiserver.go
thetechnick: While it's not big deal, but there is just no reason to setup an http handler.
As we don't start a blocking manager here, a simple blocking `for` loop would have sufficed.

For me it's fine to merge this as is.

pkg/operator/internal/controllers/api-server.go
thetechnick: There are always ways to break stuff and we really need to write unittests for the operator controllers, but I don't see a need to add a test for the right import.

pkg/operator/internal/controllers/api-server.go
```diff
@@ -0,0 +1,161 @@
+/*
```
jiachengxu, line 1: The name of this file should be`apiserver_controller.go`

pkg/operator/internal/controllers/api-server.go
```diff
@@ -0,0 +1,161 @@
...
+
+	if err := r.updateStatus(ctx, apiServer, deploymentIsReady); err != nil {
+		return ctrl.Result{}, err
+	}
+
+	return ctrl.Result{}, nil
+
+}
+
+func (r *APIServerReconciler) GetDeletionObjectTypes() []runtime.Object {
```
jiachengxu, line 103: Please remove this function since it is not used anywhere

PR #389, opened 2020-04-15
Refactor e2e tests to use e2e operator

jiachengxu:
This PR refactors e2e tests to use e2e operator.
Fixes #

**Documentation**:

```release-note
NONE
```


pkg/testutil/framework.go
```diff
@@ -307,6 +315,182 @@ func (f *Framework) SetupServiceCluster(ctx context.Context, cl *RecordingClient
...
+		&rbacv1.Role{
+			ObjectMeta: metav1.ObjectMeta{
+				Name:      fmt.Sprintf("%s-leader-election-role", namePrefix),
+				Namespace: namespaceName,
+			},
+		},
+	})
+}
+
+func (f *Framework) componentCheck(
```
nmiculinic, line 478: why does this function have Framework as pointer receiver when it's unused? It could very well stand on its own.

pkg/testutil/framework.go
```diff
@@ -307,6 +315,182 @@ func (f *Framework) SetupServiceCluster(ctx context.Context, cl *RecordingClient
...
+		},
+		&apiextensionsv1.CustomResourceDefinition{
+			ObjectMeta: metav1.ObjectMeta{
+				Name: "dbs.fake.kubecarrier.io",
+			},
+		},
+	})
+}
+
+func (f *Framework) KubeCarrierCheck(
```
nmiculinic, line 421: This function doesn't utilize f Framework as pointer receiver (if we ignore `componentCheck` which doesn't utilize f at all). Thus it can be a free-standing function.

pkg/testutil/framework.go
```diff
@@ -307,6 +315,182 @@ func (f *Framework) SetupServiceCluster(ctx context.Context, cl *RecordingClient
...
+		},
+		&apiextensionsv1.CustomResourceDefinition{
+			ObjectMeta: metav1.ObjectMeta{
+				Name: "accounts.catalog.kubecarrier.io",
+			},
+		},
+	})
+}
+
+func (f *Framework) E2EOperatorCheck(
```
nmiculinic, line 380: The same objection as in KubecarrierCheck. Also as far as I can see this is only called in a single place --> installation/e2e-operator; thus it would be better inlining it there instead of bloating this file even more. The installation/e2e-operator has 10lines...this file is become humongous.

test/installation/kubecarrier.go
```diff
@@ -0,0 +1,67 @@
...
+
+		c := exec.CommandContext(ctx, "kubectl", "kubecarrier", "setup", "--kubeconfig", f.Config().ManagementExternalKubeconfigPath)
+		out, err := c.CombinedOutput()
+		t.Log(string(out))
+		require.NoError(t, err)
+
+		managementClient, err := f.ManagementClient(t)
+		require.NoError(t, err, "creating management client")
+
+		f.KubeCarrierOperatorCheck(ctx, t, managementClient, f.ManagementScheme)
```
nmiculinic, line 46: Inline this check, it is probably only called here.

pkg/testutil/framework.go
jiachengxu: There are serval functions in the `framework` package which have it as pointer receiver and don't use it, I wanted to have that so it can keep in the context of `framework`.

pkg/testutil/framework.go
jiachengxu: For inline `KubeCarrierCheck` and other check functions:
I keep them in the framework package, because they will be invoked multiple times once we have more service clusters, and kubecarrier installations in our e2e tests, and I didn't see much benefits to inline them so I just keep them in the framework package, and we can reuse them later.

pkg/testutil/framework.go
```diff
@@ -307,6 +315,182 @@ func (f *Framework) SetupServiceCluster(ctx context.Context, cl *RecordingClient
...
 }
 
+func (f *Framework) KubeCarrierOperatorCheck(
+	ctx context.Context,
+	t *testing.T,
+	managementClient *RecordingClient,
+	managementScheme *runtime.Scheme,
+) {
+	namePrefix := constants.KubeCarrierDefaultName + "-operator"
+	namespaceName := constants.KubeCarrierDefaultNamespace
```
jiachengxu, line 325: @nmiculinic also, putting check functions in the framework can help we use the constants from the internal/constants packages, otherwise, we need to hardcode these constants in the e2e tests, having defaulting constants in 2 places is not really nice

nmiculinic: hmm....ok; that's a valid reason to stay withing `testutil`, but not withing framework. Having it a free function without framework seems a nicer solution long term.

nmiculinic:
I've re-reviewed the PR, read the code, and read my comments. 

I'd still prefer:

* KubeCarrierOperatorCheck
* KubeCarrierCheck
* KubeCarrierOperatorCheck

be defined close to their usage since they are single-use functions in the PR and not having a framework pointer receiver. I like that logic is extracted in the function.

In the future, if there's a need, we can easily refactor the code differently.

jiachengxu:
> I've re-reviewed the PR, read the code, and read my comments.
> 
> I'd still prefer:
> 
> * KubeCarrierOperatorCheck
> * KubeCarrierCheck
> * KubeCarrierOperatorCheck
> 
> be defined close to their usage since they are single-use functions in the PR and not having a framework pointer receiver. I like that logic is extracted in the function.
> 
> In the future, if there's a need, we can easily refactor the code differently.

I put check functions in the `testutil` without having framework as pointer receiver. I didn't put the functions directly in the installation tests because then we need to hardcode these constants(`KubeCarrierDefaultName` and `KubeCarrierDefaultNamespace`) in the e2e tests because they are in internal/constants package which can not be used in the test folder, having defaulting constants in 2 places is not really nice
See decision here for more details: https://github.com/kubermatic/kubecarrier/pull/389#discussion_r416383571

PR #398, opened 2020-04-22
Init grpc

nmiculinic:
Implements initial gRPC support for the APIServer

Fixes #366 

```release-note
NONE
```


config/internal/api-server/manager/service.yaml
```diff
@@ -0,0 +1,16 @@
+apiVersion: v1
+kind: Service
+metadata:
+  name: manager
+  namespace: system
+  labels:
+    control-plane: api-server
+spec:
+  type: "ClusterIP"
```
jiachengxu, line 9: You don't need to specify this since `ClusterIP` is the default value for Service

pkg/internal/resources/api-server/apiserver.go
```diff
@@ -32,8 +35,11 @@ import (
 type Config struct {
 	// Namespace is the KubeCarrier master controller manager should be deployed into.
 	Namespace string
-	// Name of this Kubecarrier API object.
+	// Name of this KubeCarrier API object.
 	Name string
+
+	// Spec of the APIServer
+	Spec operatorv1alpha1.APIServerSpec
```
jiachengxu, line 42: Why don't you just pass the secret name here?

hack/proto-codegen.sh
```diff
@@ -0,0 +1,54 @@
...
+for pkg in ${PBUFS} ; do
+  abs_path=${PROJECT}/${pkg}
+  echo Generating from '*.proto' in $abs_path
+  protoc \
+    --go_out=plugins=grpc:${abs_path}  \
+    --grpc-gateway_out=logtostderr=true:${abs_path} \
+    --swagger_out=logtostderr=true:${abs_path} \
+    -I${PROJECT}/bin/protoc-bin/include \
+    -I${GOPATH}/pkg/mod/github.com/grpc-ecosystem/grpc-gateway@v1.14.3/third_party/googleapis \
+    -I/usr/include \
```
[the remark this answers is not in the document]
nmiculinic: I like system wide protoc installation more.

pkg/internal/resources/api-server/apiserver.go
nmiculinic: Because many other stuff shall be needed in the future and I'll rather pass the whole spec instead of writing glue code which translates stuff around

hack/proto-codegen.sh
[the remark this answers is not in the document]
jiachengxu: but then why do you need `-I${PROJECT}/bin/protoc-bin/include \`?

pkg/apis/apiserver/v1alpha1/doc.go
```diff
@@ -0,0 +1,18 @@
...
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+// v1alpha1 API definitions for KubeCarrier
+package v1alpha1
```
thetechnick, line 18: I would like to start with just `v1`, while the whole project is in alpha, also our APIs need to be considered alpha and I don't want to support multiple `alpha`, `beta`, etc api versions. If we need to do breaking API changes we will bump our major version.

In difference to the kube-apiserver, new apiversions here don't come with no cost overhead, so we should keep it simple.

pkg/apis/apiserver/v1alpha1/kubecarrier.proto
```diff
@@ -0,0 +1,43 @@
...
+  google.protobuf.Timestamp buildDate = 3;
+  string goVersion = 4;
+  string platform = 5;
+}
+
+// Requests/Responses
+message VersionRequest {}
+
+// Service
+service Kubecarrier {
```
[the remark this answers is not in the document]
thetechnick: `KubeCarrier` is ok for me, as we also need to implement probe (e.g. readiness) checks, somewhere and we don't need to have multiple services for it - as the logic _should_ be somewhat simple.

Although at the current time just calling it just `Version` would make more sense, as that is all that it's doing and we should not speculate about implementations in future PRs...

pkg/api-server/handler.go
```diff
@@ -0,0 +1,44 @@
...
+import (
+	"context"
+
+	"github.com/golang/protobuf/ptypes/timestamp"
+
+	"github.com/kubermatic/kubecarrier/pkg/apis/apiserver/v1alpha1"
+	"github.com/kubermatic/kubecarrier/pkg/internal/version"
+)
+
+type kubecarrierHandler struct{}
```
thetechnick, line 28: Please call this KubeCarrierServer or VersionServer (ref. https://github.com/kubermatic/kubecarrier/pull/398/files#r413441230)

I think Handler is missleading as it's not a normal http handler, but the implementation of the `KubeCarrierServer` interface.

pkg/apis/operator/v1alpha1/apiserver_types.go
```diff
@@ -22,6 +22,8 @@ import (
 
 // APIServerSpec defines the desired state of APIServer
 type APIServerSpec struct {
+	// TLSSecretRef referencing the TLS serving credentials for the API server
```
thetechnick, line 25:
```suggestion
	// TLSSecretRef references the TLS certificate and private key for serving the KubeCarrier API.
```
I wanted to get rid of `credentials` here, as TLS credentials are usually things like client certs and I don't want any confusion here.

pkg/internal/resources/api-server/apiserver.go
thetechnick: Please add the glue code here.

I would like to have the code in here just executing the template, while the controller needs to prepare/default fields as needed. Coupling this directly to the API structs is just lazy.

pkg/internal/util/cmd.go
```diff
@@ -30,7 +30,7 @@ import (
 // CmdLogMixin adds necessary CLI flags for logging and setups the controller runtime log
 func CmdLogMixin(cmd *cobra.Command) *cobra.Command {
 	dev := cmd.PersistentFlags().Bool("development", terminal.IsTerminal(int(os.Stdout.Fd())), "format output for console")
-	v := cmd.PersistentFlags().Int8P("verbose", "v", 0, "verbosity level")
+	v := cmd.PersistentFlags().Int8P("verbose", "v", 4, "verbosity level")
```
thetechnick, line 33: So this set's the default verbosity to level 4 right?
Don't think we want to have debug logging enabled by default.

test/integration/apiserver.go
```diff
@@ -43,11 +57,116 @@ func newAPIServer(f *testutil.Framework) func(t *testing.T) {
...
+					t.Log("gRPC server temporary unavailable, retrying")
+					return false, nil
+				}
+			}
+			return false, err
+		}, versionCtx.Done()), "client version gRPC call")
 	}
 }
+
+type ToGRPCStatus interface {
```
thetechnick, line 170:
```suggestion
type toGRPCStatus interface {
```
should be private

pkg/apis/apiserver/v1alpha1/kubecarrier.proto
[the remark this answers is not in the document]
nmiculinic: Because this is intended for containing all our `v1alpha1` API endpoints

pkg/api-server/apiserver.go
```diff
@@ -54,21 +62,62 @@ func NewAPIServer() *cobra.Command {
...
+	}
+
+	server := http.Server{
+		Handler: handlerFunc,
+		Addr:    flags.addr,
+	}
+
+	log.Info("booting serving API-server", "addr", flags.addr)
+	if flags.TLSCertFile == "" {
+		log.V(4).Info("No TLS cert file defined, skipping TLS setup")
```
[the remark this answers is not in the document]
nmiculinic: Grpcweb is working without TLS, and it's easier to test locally without it.

hack/lib.sh
```diff
@@ -0,0 +1,21 @@
+#!/usr/bin/env bash
```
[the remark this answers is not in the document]
nmiculinic: It's intended to be generic, for it to be included all the things that's reusable. I really don't want to include multiple files in bash with its non-existent dependency management system.

pkg/apis/apiserver/v1alpha1/doc.go
nmiculinic: what's the cost overhead you speak off? the `v1alpha1` makes a pretty nice semantic how stable this API is, and it's something I really enjoy regarding the API versioning from kubernetes.

pkg/internal/util/cmd.go
nmiculinic: In the syslog, 4 is info level, 6 is the debug level. Thus we should honor the syslog conventions.

pkg/api-server/apiserver.go
```diff
@@ -40,35 +47,78 @@ func init() {
 }
 
 type flags struct {
-	port int
+	addr              string
```
aborilov, line 50: this should be `address`

hack/lib.sh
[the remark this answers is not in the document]
thetechnick: Please move these variables into the Makefile and remove this bash include for now.
All our scripts are intended to be only called via make and we should not start putting our variables in multiple places.

If we need shared bash code in multiple scripts, we can revisit such a `lib.sh`, although I would rather have longer scripts with different command args than using includes in bash.

pkg/api-server/apiserver.go
nmiculinic: done

hack/lib.sh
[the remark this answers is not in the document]
nmiculinic: It's included in two places, and I'd rather have those things defined in one place:

* test.Dockerfile building
* proto generation

I don't want c/p the constants around without the good need for it.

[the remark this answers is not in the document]
nmiculinic: The `testDockerfile` isn't built via calling the make (internally, it just c/p a few bash files and runs them) since it would entail copying bunch of files and recreating the folder structure the make expects...and I'd rather depend on a few files as possible while building the `test.Dockerfile`

pkg/api-server/api/v1/doc.go
```diff
@@ -0,0 +1,18 @@
+/*
```
jiachengxu, line 1: please rename `api-server` folder to `apiserver` as we discussed

pkg/api-server/internal/v1/handler.go
```diff
@@ -0,0 +1,44 @@
+/*
```
jiachengxu, line 1: please rename this file to kubecarrier.go

jiachengxu: and please move this file under `services` folder, because there will be `authorizer` folder also under internal.

nmiculinic: done

pkg/api-server/api/v1/doc.go
nmiculinic: done

hack/proto-codegen.sh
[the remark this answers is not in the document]
thetechnick: @nmiculinic why was this marked as resolved?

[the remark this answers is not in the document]
nmiculinic: see `Makefile` ; this is used within the `quay.io/kubecarrier/test` container since it's the easier way for its replicability. Inside the container the protoc is installed with the proper version under `/usr/include`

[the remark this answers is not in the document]
nmiculinic: also, there's a comment at file's beginning.

PR #401, opened 2020-04-27
Authorization

jiachengxu:

Fixes #

**Documentation**:

```release-note
NONE
```


nmiculinic:
Why do we need this exactly? Why not just perform the operation in the users name and let the k8s handle if it's allowed or not?

pkg/api-server/apiserver.go
```diff
@@ -41,6 +41,7 @@ import (
 	apiservicev1 "github.com/kubermatic/kubecarrier/pkg/api/v1/services"
 	apiserverv1alpha1 "github.com/kubermatic/kubecarrier/pkg/apis/apiserver/v1alpha1"
 	catalogv1alpha1 "github.com/kubermatic/kubecarrier/pkg/apis/catalog/v1alpha1"
+	authorizer2 "github.com/kubermatic/kubecarrier/pkg/internal/authorizer"
```
nmiculinic, line 44: how bout better name than `authorizer2`

pkg/internal/authorizer/authorizer.go
```diff
@@ -0,0 +1,108 @@
...
+	}
+	if err := a.Create(ctx, review); err != nil {
+		return DecisionDenied, fmt.Errorf("creating SubjectAccessReview: %s", err)
+	}
+	switch {
+	case review.Status.Denied:
+		return DecisionDenied, nil
+	case review.Status.Allowed:
+		return DecisionAllowed, nil
+	default:
```
nmiculinic, line 105: When is this possible? The default should be Denied for security purposes....what the client should do if there's no opinion?

jiachengxu: the client only start operations if the Decision is DecisionAllowed, the `NoOpinion` is used for providing more detailed info.

pkg/api/v1/services/offering.go
```diff
@@ -26,19 +26,21 @@ import (
 
 	v1 "github.com/kubermatic/kubecarrier/pkg/api/v1"
 	catalogv1alpha1 "github.com/kubermatic/kubecarrier/pkg/apis/catalog/v1alpha1"
+	"github.com/kubermatic/kubecarrier/pkg/internal/authorizer"
 )
 
 type offeringServer struct {
-	client client.Client
+	authorizationClient authorizer.AuthorizationClient
```
aborilov, line 33: we can leave it as `client.Client`

pkg/internal/authorizer/authorizer.go
```diff
@@ -0,0 +1,172 @@
...
+func (a AuthorizationClient) Get(ctx context.Context, key client.ObjectKey, obj runtime.Object) error {
+	decision, err := a.authorize(ctx, obj, GetAuthorization{
+		Name:      key.Name,
+		Namespace: key.Namespace,
+	})
+	if err != nil {
+		return fmt.Errorf("authorizing request: %w", err)
+	}
+	if decision != DecisionAllowed {
+		return fmt.Errorf("not allowed: %w", err)
```
nmiculinic, line 53:
```suggestion
		return fmt.Errorf("forbidden: %w", err)
```

in line with HTTP 403

pkg/internal/authorizer/authorizer.go
```diff
@@ -0,0 +1,172 @@
...
+type ListAuthorization struct {
+	Namespace string
+}
+
+func (a ListAuthorization) apply(review *authv1.SubjectAccessReview) {
+	review.Spec.ResourceAttributes.Verb = string(RequestList)
+	review.Spec.ResourceAttributes.Namespace = a.Namespace
+}
+
+func (a AuthorizationClient) getUserInfo(ctx context.Context) (UserInfo, error) {
```
nmiculinic, line 160: Please don't mix member function ordering for structs. Keep it sorted.

pkg/internal/authorizer/authorizer.go
```diff
@@ -0,0 +1,172 @@
...
+	if decision != DecisionAllowed {
+		return fmt.Errorf("not allowed: %w", err)
+	}
+	return a.Client.Get(ctx, key, obj)
+}
+
+func (a AuthorizationClient) List(ctx context.Context, list runtime.Object, opts ...client.ListOption) error {
+	var authOpt ListAuthorization
+	for _, opt := range opts {
+		if namespace, isInNamespace := opt.(client.InNamespace); isInNamespace {
```
nmiculinic, line 61: This won't work if you have some custom option setting the namespace. Could you instead run all those `client.ListOption` on `metav1.ListOption` IIRC and check the resulting object if the namespace is setup?

pkg/internal/authorizer/authorizer.go
```diff
@@ -0,0 +1,172 @@
...
+)
+
+const (
+	userHeaderKey = "user"
+)
+
+type AuthorizationClient struct {
+	Scheme *runtime.Scheme
+	Log    logr.Logger
+	client.Client
```
nmiculinic, line 41: don't embed this field; you should implement all `client.Client` function and run authZ layer before proxying the request via the client. 

By embedding the field you're not checking authz for PATCH/UPDATE and some other verbs for example, since you haven't implemented them.

pkg/internal/authorizer/authorizer.go
```diff
@@ -0,0 +1,172 @@
...
+	DecisionDenied    AuthorizationDecision = "denied"
+	DecisionNoOpinion AuthorizationDecision = "no opinion"
+)
+
+func (a AuthorizationClient) authorize(
+	ctx context.Context,
+	objType runtime.Object,
+	option AuthorizationOption,
+) (AuthorizationDecision, error) {
+	userInfo, err := a.getUserInfo(ctx)
```
nmiculinic, line 88: This is handled by the OIDC middleware layer...why are you having your own function here?

pkg/internal/authorizer/authorizer.go
```diff
@@ -0,0 +1,172 @@
...
+		return DecisionDenied, fmt.Errorf("cannot get the user infomation from the request: %w", err)
+	}
+
+	gvk, err := apiutil.GVKForObject(objType, a.Scheme)
+	if err != nil {
+		return DecisionDenied, fmt.Errorf("cannot get GVK for %T: %w", objType, err)
+	}
+	var resource string
+	if meta.IsListType(objType) {
+		resource = flect.Pluralize(strings.ToLower(strings.TrimSuffix(gvk.Kind, "List")))
```
nmiculinic, line 99: `resource` should be calculated from the RESTMapper, not guessed by using heuristics.

pkg/apiserver/internal/authorizer/authorizer.go
```diff
@@ -0,0 +1,182 @@
...
+	"google.golang.org/grpc/status"
+	authv1 "k8s.io/api/authorization/v1"
+	"k8s.io/apimachinery/pkg/api/meta"
+	"k8s.io/apimachinery/pkg/runtime"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/client/apiutil"
+)
+
+const (
+	userHeaderKey = "authorization"
```
aborilov, line 37: now when we have oid merged, you can use `oidc.ExtractUserInfo` method

pkg/apiserver/internal/authorizer/authorizer.go
```diff
@@ -0,0 +1,182 @@
...
+)
+
+type AuthorizationClient struct {
+	Scheme *runtime.Scheme
+	Log    logr.Logger
+	client.Client
+}
+
+func (a AuthorizationClient) Get(ctx context.Context, key client.ObjectKey, obj runtime.Object) error {
```
aborilov, line 46: we also need to add `Create` and `Delete` methods for `Instances` service

pkg/internal/authorizer/authorizer.go
jiachengxu: This function is removed because we have OIDC support

pkg/internal/authorizer/authorizer.go
jiachengxu: hmm, how about 401?

pkg/internal/authorizer/authorizer.go
jiachengxu: done

pkg/internal/authorizer/authorizer.go
jiachengxu: Please check the new implementation.

pkg/apiserver/internal/authorizer/authorizer.go
jiachengxu: done

pkg/apiserver/internal/authorizer/authorizer.go
jiachengxu: Please check the new implementation, and now we don't need to implement different method, we can use the same `Authorize` method everywhere

pkg/internal/authorizer/authorizer.go
jiachengxu: Also, I have tried to implement authorizer as a Middleware (Just like our OIDC middleware), register it as `mux.MiddlewareFunc`, so we don't need to handle authorization in the request handler(Get/List, etc.), and we just do the business logic there.
But the issue is that how can I map the endpoint/ query parameters to the resources that the user wants to access? For example, for the offering list endpoint, the HTTP request will like this:
```
GET https://localhost:9443/v1/accounts/{account}/offerings
```
How can I map this to:
User wants to verb: `get`, version: `v1alpha1`, group: `catalog.kubecarrier.io`, resources: `offerings` in namespace: `account`  
Does this mean that I need to build my own `RestMapper` to handle this?
WDYT @nmiculinic @thetechnick @aborilov ?

aborilov: https://github.com/kubermatic/kubecarrier/pull/443/files
here is my proposal for that.
we can use so-called `service wrappers` that isolate business logic from validation/auth/oidc, etc.

jiachengxu: Well, we can still isolate the business logic without having the wrappers, we can just put the business logic into `handleRequest` functions, like what I did in the PR.

The issue of this Implementation is that we still need to worry about authorization in every endpoint, for example, in my PR, I need to add `Authorize` to the handler when I add new service endpoints, in your approach, we need to add new wrapper for new implemented service endpoints.

If we can implement it like our OIDC middleware, we just register it as a middleware handler, and if new endpoints are added, we don't need to do extra work to make the authorization work.

aborilov: >Well, we can still isolate the business logic without having the wrappers, we can just put the business logic into handleRequest functions, like what I did in the PR.

you already have one `handleRequest` function for `oids` and one for `authorization` and number will keep growing, it's not pleasant to have all of them in one place.

> in your approach, we need to add new wrapper for new implemented service endpoints.

a good thing about this approach, you can easily go to `service_auth` file and find which endpoint has authorization and which one not, in case we want to have endpoint without authorization(like `version`). Same for validation, you don't need to search the handler function where validation start and where business logic ends, you just jump to `service_validation` file. And those files much easier to test, you can test separately authorization and business logic.

>If we can implement it like our OIDC middleware, we just register it as a middleware handler, and if new endpoints are added, we don't need to do extra work to make the authorization work.

that will be the best solution for us. the only problem that we can have, when we want to have some special cases, like leave endpoint without authorization, will be hard to configure that in middleware

jiachengxu: > you already have one handleRequest function for oids and one for authorization and number will keep growing, it's not pleasant to have all of them in one place.

Nope, I only have one handleRequest for every handler to include business logic, the number will not grow.

aborilov: that mean, you will have put all other logic in one method

jiachengxu: For account endpoint, we will only have oidc logic to get userInfo, because we need to filter accounts by using the username.
For rest of endpoint, only authorization logic is needed.

aborilov:
```
func (o accountServer) List(ctx context.Context, req *v1.AccountListRequest) (res *v1.AccountList, err error) {
	user, present := oidc.ExtractUserInfo(ctx)
	if !present {
		return nil, status.Error(codes.Unauthenticated, "unauthenticated user")
	}
	if err := o.authorizer.Authorize(ctx, &catalogv1alpha1.Account{}, authorizer.AuthorizationOption{
		Namespace: req.Account,
		Verb:      authorizer.RequestList,
	}); err != nil {
		return nil, status.Error(codes.Unauthenticated, err.Error())
	}
	return o.handleListRequest(ctx, req, user.GetName())
```
something like that, right?
this one is hard to test because you have both authorizer and oidc in one place

jiachengxu: nope, account endpoint will not have authorization logic

aborilov: ok, then

jiachengxu: This is because we implemented the account endpoint differently, for other endpoints, we use RBAC and SubjectAccessReview to do authorization/permission validation, but for account endpoint, because user can not access accounts anyways(no rbac for them to access accounts), so we don't need to authorize it, we just filter accounts according to accout.Spec.Subject

aborilov: but I still don't like to unit test some `private` method `handleListRequest` instead of the main `List` method

jiachengxu: > but I still don't like to unit test some `private` method `handleListRequest` instead of the main `List` method

Agree, but the logic we are testing in this PR and your PR are the same right :) ? I mean in this PR, I am testing provider `handleRequest` function, and in your PR, you just test `List` function, and you didn't test wrapper `List` function, right? The number of covered lines should be the same I guess :)

aborilov: correct, but the idea of this wrapper, that if I wrote tests on business logic and business logic doesn't change, I don't need to change tests. In your case, you have to switch unit tests to the private method. but that also ok because you need to do this only once. but then, if you will decide to test authorization, you can do this without call business logic. also, if you add some other logic to `List` method you will again have to fix authorization tests. 
this approach can be very helpful when we will decide to add `audit`, we can simply add new wrapper with audit logic and don't touch any other code

jiachengxu: >  but then, if you will decide to test authorization, you can do this without call business logic. also, if you add some other logic to List method you will again have to fix authorization tests.
For this one, I don't really know how to unit test authorization stuff, so I would use e2e tests for it, also, if I want to unit test authorization, I will add test to the `authorizer` middleware folder, right? so I don't need to fix authorization tests when I add other logic to `List` method

PR #411, opened 2020-05-07
[API Server] Provider service

aborilov:

Fixes #

**Documentation**:

```release-note
NONE
```


pkg/apiserver/api/v1/provider.proto
```diff
@@ -0,0 +1,66 @@
...
+
+message AccountMetadata {
+  string displayName = 1;
+  string description = 2;
+}
+
+service ProviderService {
+  rpc List(ProviderListRequest) returns (ProviderList) {
+    option (google.api.http) = {
+      get : "/v1/tenants/{tenant}/providers"
```
nmiculinic, line 58:
```suggestion
      get : "/v1/providers"
```

pkg/apiserver/api/v1/provider.proto
```diff
@@ -0,0 +1,66 @@
...
+    };
+  };
+  rpc Get(ProviderRequest) returns (Provider) {
+    option (google.api.http) = {
+      get : "/v1/tenants/{tenant}/providers/{name}"
```
nmiculinic, line 63:
```suggestion
      get : "/v1/providers/{name}"
```

pkg/apiserver/api/v1/provider.proto
jiachengxu: nope, this is the endpoint for listing provider objects in the account namespace, so it should be:
"/v1/accounts/{account}/providers"

jiachengxu: and also for the following "get"

PR #412, opened 2020-05-07
[API Server] Offering service

jiachengxu:
This PR implements the offering service to get/list offerings objects.
Fixes #

**Documentation**:

```release-note
NONE
```


pkg/apiserver/internal/v1/offering.go
```diff
@@ -0,0 +1,146 @@
...
+			Selector: selector,
+		})
+	}
+	if req.Continue != "" {
+		listOptions = append(listOptions, client.Continue(req.Continue))
+	}
+	return listOptions, nil
+}
+
+func (o offeringServer) validateGetRequest(req *v1.OfferingRequest) error {
```
aborilov, line 107: because we are putting all service in the same `v1` package it's better to name this method something like `validateOfferingRequest`, in that case we can use it in other services that need to use `offering`

pkg/apiserver/internal/v1/offering.go
```diff
@@ -0,0 +1,146 @@
...
+		Name:      req.Name,
+		Namespace: req.Tenant,
+	}, offering); err != nil {
+		return nil, status.Errorf(codes.Internal, fmt.Sprintf("getting offering: %s", err.Error()))
+	}
+	return o.convertOffering(offering), nil
+
+}
+
+func (o offeringServer) validateListRequest(req *v1.OfferingListRequest) ([]client.ListOption, error) {
```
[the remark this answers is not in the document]
jiachengxu: What is the benefit to split logic, I tried first to split logic first, but I just realized that I still need to do some logic in the validation step, for example, I need to parse the labelSelector twice, once in the validation, once in the extracting of listoptions, that's why I put them together.

pkg/apiserver/internal/v1/offering.go
jiachengxu: 1. the receiver of this function is `offeringServer` so naming it to `validateOfferingRequest` is a bit redundant.
> in that case we can use it in other services that need to use offering
2. I don't really understand this, why other services will use `OfferingRequest`? Can you please provide an example here?

pkg/apiserver/internal/v1/offering.go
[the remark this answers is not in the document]
aborilov: >I need to parse the labelSelector twice
I see but I don't think that's a big deal
currently, the name of the method doesn't' match what it's doing.
what do you think about that variant?
https://github.com/kubermatic/kubecarrier/blob/1f53bdb625545f084f65526ed5b6f40bf12a8a8f/pkg/apiserver/internal/v1/provider.go#L48-L78

pkg/apiserver/internal/v1/offering.go
aborilov: > 1. the receiver of this function is `offeringServer` so naming it to `validateOfferingRequest` is a bit redundant.

yeah, I was thinking about decoupling those methods from `offeringServer` because they don't really need it at all 

but, meh, I think you are right, let's make it simple

pkg/apiserver/internal/v1/offering.go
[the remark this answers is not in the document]
jiachengxu: To be honest, I don't really see the benefits to split it up into 2 functions. In my opinion, the `validateListRequest` is doing validation as following:
1. validate the query parameter
2. if the parameter is valid, then put it into listOptions
3. if the parameter is not valid, return error.
I don't really understand why you want to split the only part.2 out. I think it is fine to add listoptions while doing validation.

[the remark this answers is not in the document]
jiachengxu: Also in your example, you didn't really map the query parameters to the listoptions, for example, every time you want to add another query parameter, you need to add it in the `validateProviderListRequest` and then append listoptions out of the function (in a different place), which means you need to sync these 2 places manually, but if you put them together, appending listoption just after validating the parameter, you don't need to do that, and the flow is easy to follow and understand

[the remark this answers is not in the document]
aborilov: agree, I'm ok with both ways. what do you think about renaming it to `getListOptions`?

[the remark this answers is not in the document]
jiachengxu: both are working for me, but I would prefer the `validateListRequest`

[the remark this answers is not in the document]
aborilov: but it's not consistent with `validateGetRequest`, which only validates and returns error if request not valid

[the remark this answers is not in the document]
jiachengxu: but if you rename it to `getListOptions` it is still not consistent with Get Request, right :) ?
for Get request, you have a validation function, but for List request, you don't have one :)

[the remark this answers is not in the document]
aborilov: yes, this method not a part of interface, so you can have it or not. I'm trying to say, that `validation` method should return an error or not. if you still think that's ok to return values in `validate` method, I can live with it :)

pkg/apiserver/api/v1/offering.proto
```diff
@@ -0,0 +1,67 @@
...
+  ObjectReference provider = 4;
+  CRDInformation crd = 5;
+}
+
+message OfferingList {
+  repeated Offering items = 1;
+  string continue = 2;
+}
+
+message OfferingRequest {
```
thetechnick, line 38:
```suggestion
message OfferingGetRequest {
```

pkg/apiserver/api/v1/offering.proto
```diff
@@ -0,0 +1,67 @@
...
+syntax = "proto3";
+package kubecarrier.api.v1;
+option go_package = "v1";
+
+import "google/api/annotations.proto";
+
+import "types.proto";
+
+message Offering {
+  string name = 1;
```
thetechnick, line 26: I would like to put name, labels and annotations into our own common ObjectMeta object, similar to Kubernetes.
This common metadata object makes generic handler functions of metadata stuff much easier to implement :)
(And we don't have to add the same fields to EVERY object)

e.g.
```proto
message ObjectMeta {
  string uid = 1; // important to know wheter it's still the same instance we are dealing with or if an object with the same name has been created
  string name = 2;
  google.protobuf.Timestamp created = 3; // can also be creationTimestamp to align with Kubernetes
  google.protobuf.Timestamp deleted = 4; // can also be deletionTimestamp to align with Kubernetes
  string resourceVersion = 5; // super important for concurrency control handling
  map<string, string> labels = 6; // we support label-selectors, so we also need to display them
  map<string, string> annotations = 7; // always nice-to-have as a easy and quick extension point
  uint64 generation = 8; // wether spec/metadata has changed
}
```
this common Metadata should be defined in it's own `meta.proto` and imported here.

In the future we might also want to add and expose `GenerateName` and `Finalizers`.

Other fields should also be put into `Spec` or `Status` sub-objects following the Kubernetes setup.

pkg/apiserver/api/v1/offering.proto
```diff
@@ -0,0 +1,67 @@
...
+  string tenant = 2;
+  OfferingMetadata metadata = 3;
+  ObjectReference provider = 4;
+  CRDInformation crd = 5;
+}
+
+message OfferingList {
+  repeated Offering items = 1;
+  string continue = 2;
```
thetechnick, line 35: Same as with common `ObjectMeta`, I would like to have a `ListMeta` object which contains `continue` and `resourceVersion` fields.
e.g. https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.17/#listmeta-v1-meta

pkg/apiserver/api/v1/offering.proto
```diff
@@ -0,0 +1,67 @@
...
+
+service OfferingService {
+  rpc List(OfferingListRequest) returns (OfferingList) {
+    option (google.api.http) = {
+      get : "/v1/tenants/{tenant}/offerings"
+    };
+  };
+  rpc Get(OfferingRequest) returns (Offering) {
+    option (google.api.http) = {
+      get : "/v1/tenants/{tenant}/offerings/{name}"
```
thetechnick, line 64: As the underlying object is the `Account`, it probably makes more sense to name it `/v1/account/{account}/*`?
This would also translate into our `ObjectMeta` object having an `account` field instead of a `namespace` in Kubernetes.

WDYT @aborilov @jiachengxu @nmiculinic ?

pkg/apiserver/api/v1/offering.proto
```diff
@@ -0,0 +1,67 @@
...
+
+import "google/api/annotations.proto";
+
+import "types.proto";
+
+message Offering {
+  string name = 1;
+  string tenant = 2;
+  OfferingMetadata metadata = 3;
+  ObjectReference provider = 4;
```
nmiculinic, line 29: This doesn't feel right; why cannot this be a string? It makes easier API usage, more straightforward. Same as with tenant.

pkg/apiserver/api/v1/offering.proto
nmiculinic: agreed

test/integration/apiserver.go
```diff
@@ -164,9 +166,179 @@ func newAPIServer(f *testutil.Framework) func(t *testing.T) {
...
+				},
+				Crd: &apiserverv1.CRDInformation{
+					Name:     "test-crd",
+					ApiGroup: "test-crd-group",
+					Kind:     "test-kind",
+					Plural:   "test-plural",
+					Versions: []*apiserverv1.CRDVersion{
+						{
+							Name:   "test-version",
+							Schema: "{\"openAPIV3Schema\":{\"type\":\"object\",\"properties\":{\"apiVersion\":{\"type\":\"string\"}}}}",
```
nmiculinic, line 331: you can use golang raw string here. https://golang.org/ref/spec#String_literals

pkg/apiserver/api/v1/offering.proto
jiachengxu: I just want to keep consistent with our catalogv1alpha1.Offering definition

pkg/apiserver/api/v1/meta.proto
```diff
@@ -0,0 +1,37 @@
...
+
+message ObjectMeta {
+  string uid = 1;
+  string name = 2;
+  google.protobuf.Timestamp creationTimestamp = 3;
+  google.protobuf.Timestamp deletionTimestamp = 4;
+  string resourceVersion = 5;
+  map<string, string> labels = 6;
+  map<string, string> annotations = 7;
+  int64 generation = 8;
```
jiachengxu, line 31: I use `int64` instead of `uint64` since kubernetes uses `int64` and want to keep consistent with it.

test/integration/apiserver.go
jiachengxu: thanks！

pkg/apiserver/api/v1/offering.proto
```diff
@@ -0,0 +1,70 @@
...
+package kubecarrier.api.v1;
+option go_package = "v1";
+
+import "google/api/annotations.proto";
+
+import "types.proto";
+import "meta.proto";
+
+message Offering {
+  ObjectMeta objectMeta = 1;
```
thetechnick, line 27:
```suggestion
  ObjectMeta metadata = 1;
```
just to keep in line with k8s

pkg/apiserver/api/v1/meta.proto
thetechnick: :+1: Could you also please add a `string account = 9;`?
As this field is essentially our `Namespace`

pkg/apiserver/api/v1/offering.proto
jiachengxu: I have also changed the `listMeta` to `metadata` in OfferingList :)

PR #421, opened 2020-05-13
Offering Watch API

jiachengxu:

Fixes #

**Documentation**:

```release-note
NONE
```


pkg/apiserver/internal/v1/offering.go
```diff
@@ -21,67 +21,133 @@ import (
...
+func (o offeringServer) Watch(req *v1.OfferingWatchRequest, stream v1.OfferingService_WatchServer) error {
+	listOptions, err := o.validateWatchRequest(req)
+	if err != nil {
+		return status.Error(codes.InvalidArgument, err.Error())
+	}
+	watcher, err := o.dynamicClient.Resource(o.gvr).Namespace(req.Account).Watch(listOptions)
+	if err != nil {
+		return status.Errorf(codes.Internal, "watching offerings: %s", err.Error())
+	}
+	for {
```
nmiculinic, line 117: `defer watcher.Stop()` to prevent resource leakage.

pkg/apiserver/api/v1/offering.proto
```diff
@@ -70,4 +78,9 @@ service OfferingService {
       get : "/v1/accounts/{account}/offerings/{name}"
     };
   };
+  rpc Watch(OfferingWatchRequest) returns (stream Event) {
```
nmiculinic, line 81: I dislike this untyped return; Have a typed return instead.

pkg/apiserver/api/v1/event.proto
```diff
@@ -0,0 +1,26 @@
...
+
+syntax = "proto3";
+package kubecarrier.api.v1;
+option go_package = "v1";
+
+import "google/protobuf/any.proto";
+
+message Event {
+  google.protobuf.Any object = 1;
+  string type = 2;
```
nmiculinic, line 25: https://developers.google.com/protocol-buffers/docs/proto3#enum

use enum for Event type (and as I've said in later comment, this should be typed, not untyped)

pkg/apiserver/apiserver.go
```diff
@@ -117,14 +122,38 @@ func runE(flags *flags, log logr.Logger) error {
...
+	certPool.AppendCertsFromPEM(certs)
+	grpcTLSConfig := &tls.Config{
+		InsecureSkipVerify: true,
+		RootCAs:            certPool,
+	}
+	// Create grpc client for watch api
+	grpcClient, err := grpc.DialContext(ctx, flags.address, grpc.WithTransportCredentials(credentials.NewTLS(grpcTLSConfig)))
+	if err != nil {
+		return err
+	}
```
nmiculinic, line 145: extract this into `createInternalGRPCClient` method.

pkg/apiserver/apiserver.go
```diff
@@ -117,14 +122,38 @@ func runE(flags *flags, log logr.Logger) error {
...
+
+	ctx := context.Background()
+	certPool := x509.NewCertPool()
+	certs, err := ioutil.ReadFile(flags.TLSCertFile)
+	if err != nil {
+		return err
+	}
+	certPool.AppendCertsFromPEM(certs)
+	grpcTLSConfig := &tls.Config{
```
nmiculinic, line 137: this can be inlined, no need for creating this variable.

pkg/apiserver/internal/v1/offering.go
```diff
@@ -21,67 +21,133 @@ import (
...
+			}
+			offering, err := o.convertOffering(catalogOffering)
+			if err != nil {
+				return status.Errorf(codes.Internal, "converting Offering: %s", err.Error())
+			}
+			any, err := ptypes.MarshalAny(offering)
+			if err != nil {
+				return status.Errorf(codes.Internal, "marshalling Offering to Any: %s", err.Error())
+			}
+			if err := stream.Send(&v1.Event{
```
nmiculinic, line 141: As I've already mentioned, this should be a typed response, not any.

pkg/apiserver/internal/v1/offering.go
```diff
@@ -21,67 +21,133 @@ import (
...
+		return status.Error(codes.InvalidArgument, err.Error())
+	}
+	watcher, err := o.dynamicClient.Resource(o.gvr).Namespace(req.Account).Watch(listOptions)
+	if err != nil {
+		return status.Errorf(codes.Internal, "watching offerings: %s", err.Error())
+	}
+	for {
+		select {
+		case <-stream.Context().Done():
+			return status.Error(codes.Internal, stream.Context().Err().Error())
```
nmiculinic, line 120:
```suggestion
			return stream.Context().Err()
```

this would utilize the right error code, from the documentation:

```
	// Done returns a channel that's closed when work done on behalf of this
	// context should be canceled. Done may return nil if this context can
	// never be canceled. Successive calls to Done return the same value.
	// The close of the Done channel may happen asynchronously,
	// after the cancel function returns.
	//
	// WithCancel arranges for Done to be closed when cancel is called;
	// WithDeadline arranges for Done to be closed when the deadline
	// expires; WithTimeout arranges for Done to be closed when the timeout
	// elapses.
	//
	// Done is provided for use in select statements:
	//
	//  // Stream generates values with DoSomething and sends them to out
	//  // until DoSomething returns an error or ctx.Done is closed.
	//  func Stream(ctx context.Context, out chan<- Value) error {
	//  	for {
	//  		v, err := DoSomething(ctx)
	//  		if err != nil {
	//  			return err
	//  		}
	//  		select {
	//  		case <-ctx.Done():
	//  			return ctx.Err()
	//  		case out <- v:
	//  		}
	//  	}
	//  }
	//
	// See https://blog.golang.org/pipelines for more examples of how to use
	// a Done channel for cancellation.
	Done() <-chan struct{}
```

Thus if the user cancels the watch, the error code should be `codes.Cancelled = 1` which this approach satisfies.

pkg/apiserver/internal/v1/offering.go
```diff
@@ -21,67 +21,133 @@ import (
...
+			}
+			any, err := ptypes.MarshalAny(offering)
+			if err != nil {
+				return status.Errorf(codes.Internal, "marshalling Offering to Any: %s", err.Error())
+			}
+			if err := stream.Send(&v1.Event{
+				Type:   string(event.Type),
+				Object: any,
+			}); err != nil {
+				return status.Errorf(codes.Internal, "sending Offering stream: %s", err.Error())
```
nmiculinic, line 145: here you should propagate the error code if the `err` satisfies the right interface, and use `codes.Internal` as a fallback.

```
type toGRPCStatus interface {
	GRPCStatus() *status.Status
}
```

test/integration/apiserver.go
```diff
@@ -471,6 +472,31 @@ func offeringService(ctx context.Context, conn *grpc.ClientConn, managementClien
...
+
+		// watch offerings
+		watchClient, err := client.Watch(offeringCtx, &apiserverv1.OfferingWatchRequest{
+			Account: testName,
+		})
+		require.NoError(t, err)
+		counter := 0
+		expectedEventNum := 2
+		assert.NoError(t, wait.PollUntil(time.Second, func() (done bool, err error) {
+			event, err := watchClient.Recv()
```
nmiculinic, line 484: This is a blocking call, thus no need for such polling complications. 

https://grpc.io/docs/tutorials/basic/go/#server-side-streaming-rpc-1

nmiculinic: Also, read the docs, both `RecvMsg` and `SendMsg` are blocking calls. The timeout is injected via the original context.

test/integration/apiserver.go
```diff
@@ -471,6 +472,31 @@ func offeringService(ctx context.Context, conn *grpc.ClientConn, managementClien
 			assert.Equal(t, expectedResult, offering)
 			return true, nil
 		}, offeringCtx.Done()))
+
+		// watch offerings
+		watchClient, err := client.Watch(offeringCtx, &apiserverv1.OfferingWatchRequest{
```
nmiculinic, line 477: testing wise I'd love seeing more events than pure `ADDED`. Maybe delte the offering to get the `DELETED` event type.

pkg/apiserver/api/v1/event.proto
jiachengxu: replied here: https://github.com/kubermatic/kubecarrier/pull/437#discussion_r432352448

nmiculinic: here enum is the right choice since it's unlikely members would change in the future, and a client should make a specific action depending on the event type.

pkg/apiserver/internal/v1/offering.go
jiachengxu: done

pkg/apiserver/apiserver.go

pkg/apiserver/apiserver.go

pkg/apiserver/internal/v1/offering.go
jiachengxu: done

pkg/apiserver/internal/v1/offering.go
jiachengxu: done

test/integration/apiserver.go
jiachengxu: done

test/integration/apiserver.go
jiachengxu: I have added more events, for example, `Deleted` and `Modified` events.

pkg/apiserver/api/v1/request.proto
```diff
@@ -29,3 +29,10 @@ message ListRequest {
   int64 limit = 3;
   string continue = 4;
 }
+
+message WatchRequest {
+  string account = 1;
+  string labelSelector = 2;
+  string resourceVersion = 3;
+  string operationType = 4;
```
thetechnick, line 37: Do we even need this filter by operation type - Kubernetes also does not support such a feature and I can't think of a use case for it.

pkg/apiserver/api/v1/event.proto
```diff
@@ -0,0 +1,26 @@
...
+limitations under the License.
+*/
+
+syntax = "proto3";
+package kubecarrier.api.v1;
+option go_package = "v1";
+
+import "google/protobuf/any.proto";
+
+message Event {
```
thetechnick, line 23: Maybe we should also call this `WatchEvent` to align with K8s?
https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.18/#watchevent-v1-meta

jiachengxu: done

pkg/apiserver/api/v1/request.proto
jiachengxu: done

PR #424, opened 2020-05-15
Add services API

aborilov:

Fixes #

**Documentation**:

```release-note
NONE
```


pkg/apis/fake/v1/db_types.go
```diff
@@ -43,6 +43,9 @@ type DBSpec struct {
 type OperationFlagType string
 
 func (o OperationFlagType) Enabled() bool {
+	if o == "" {
```
jiachengxu, line 46: Is this needed for this PR? Otherwise, please put it in a dedicated PR

pkg/apiserver/internal/util/util.go
```diff
@@ -28,3 +32,95 @@ func TimestampProto(t *metav1.Time) (*timestamp.Timestamp, error) {
 	}
 	return ptypes.TimestampProto(t.Time)
 }
+
+func FromMetav1(obj metav1.ObjectMeta) (*v1.ObjectMeta, error) {
```
jiachengxu, line 36: This function is not used anywhere, and also I think these helper functions should be versioned under `/v1` folder?
If this function is not used in this PR, please remove it, I have added a similar helper function in the refactor PR to share between all our services.

pkg/apiserver/internal/v1/service.go
```diff
@@ -0,0 +1,246 @@
...
+	}
+}
+func (o serviceServer) Create(ctx context.Context, req *v1.ServiceCreateRequest) (res *v1.Service, err error) {
+	obj := &unstructured.Unstructured{}
+
+	gvk := o.gvkFromService(req.Service, req.Version)
+	obj.SetGroupVersionKind(gvk)
+	val := map[string]interface{}{}
+	if err := json.Unmarshal([]byte(req.Spec.Spec), &val); err != nil {
+		return nil, status.Errorf(codes.Internal, "creating services: spec should by type of map[string]intreface{}")
```
jiachengxu, line 60:
```suggestion
		return nil, status.Error(codes.Internal, "creating service: spec should be type of map[string]intreface{}")
```

pkg/apiserver/internal/v1/service.go
```diff
@@ -0,0 +1,246 @@
...
+	if err := unstructured.SetNestedMap(obj.Object, val, "spec"); err != nil {
+		return nil, status.Errorf(codes.Internal, fmt.Sprintf("creating services: %s", err.Error()))
+	}
+	if err := util.SetMetadata(obj, req.Spec.Metadata); err != nil {
+		return nil, status.Errorf(codes.Internal, fmt.Sprintf("creating services: %s", err.Error()))
+	}
+	if err := o.client.Create(ctx, obj); err != nil {
+		return nil, status.Errorf(codes.Internal, fmt.Sprintf("creating services: %s", err.Error()))
+	}
+	if err = o.client.Get(ctx, types.NamespacedName{
```
jiachengxu, line 71: Why do you need to get the object after creating it?

pkg/apiserver/internal/v1/service.go
```diff
@@ -0,0 +1,246 @@
...
+		return nil, status.Errorf(codes.Internal, fmt.Sprintf("converting Service: %s", err.Error()))
+	}
+	return
+}
+
+func (o serviceServer) Delete(ctx context.Context, req *v1.ServiceDeleteRequest) (*empty.Empty, error) {
+	obj := &unstructured.Unstructured{}
+	gvk := o.gvkFromService(req.Service, req.Version)
+	obj.SetGroupVersionKind(gvk)
+	if err := o.client.Get(ctx, types.NamespacedName{
```
jiachengxu, line 137: Why do you need to `Get` before `Delete`?

pkg/internal/resources/elevator/elevator.go
```diff
@@ -263,6 +263,30 @@ func Manifests(c Config) ([]unstructured.Unstructured, error) {
...
+			Name: c.TenantPlural + "." + c.TenantGroup,
+			Labels: map[string]string{
+				"kubecarrier.io/apiserver": "true",
+			},
+		},
+		Rules: []rbacv1.PolicyRule{
+			{
+				APIGroups: []string{c.TenantGroup},
+				Resources: []string{c.TenantPlural},
+				Verbs:     []string{"create", "delete", "get", "list", "watch"},
```
jiachengxu, line 281: "watch" is not needed

pkg/apiserver/internal/v1/service.go
aborilov: thanks, didn't know that `Create` update `obj`

pkg/apiserver/internal/v1/service.go
aborilov: you're right, we don't need that

pkg/apiserver/internal/util/util.go
aborilov: done

pkg/apis/fake/v1/db_types.go
aborilov: will add in separate PR

pkg/internal/resources/elevator/elevator.go
aborilov: done

aborilov:
> Can you please add some unit tests and e2e tests for the new service?

of course but I thought we want this for the demo.
tests are my next step

thetechnick:
@aborilov I would like to standardize our naming convention, because with CRDs, Services and Offerings, there are a lot of dynamic things happening.

For instances of an offering, I would like to make sure we refer to them as "Service Instances" or just Instances in the documentation and also in the API.

So could you please rename the service and endpoint to `Instance*` / `/instance`?

pkg/apiserver/api/v1/instance.proto
```diff
@@ -0,0 +1,89 @@
...
+
+import "google/api/annotations.proto";
+import "google/protobuf/empty.proto";
+
+import "meta.proto";
+
+message Instance {
+  ObjectMeta metadata = 1;
+  string spec = 2;
+  string status = 3;
```
nmiculinic, line 29: This needs more information to be useful:
* maybe reference to the offering? it would satisfy most requirements since I could easily fetch the offering and find out the details
* instance image, provider, readiness status (boolean).

pkg/apiserver/api/v1/instance.proto
```diff
@@ -0,0 +1,89 @@
...
+
+message InstanceDeleteRequest {
+  string instance = 1;
+  string version = 2;
+  string name = 3;
+  string account = 4;
+}
+
+message InstanceListRequest {
+  string instance = 1;
```
nmiculinic, line 52: Why are you requiring instance string in the list request; what's the purpose of it?

pkg/apiserver/api/v1/instance.proto
```diff
@@ -0,0 +1,89 @@
...
+
+message InstanceGetRequest {
+  string instance = 1;
+  string version = 2;
+  string name = 3;
+  string account = 4;
+}
+
+message InstanceDeleteRequest {
+  string instance = 1;
```
nmiculinic, line 45: What's the difference between `instance` and `name`?

pkg/apiserver/api/v1/instance.proto
```diff
@@ -0,0 +1,89 @@
...
+  string account = 3;
+  string labelSelector = 4;
+  int64 limit = 5;
+  string continue = 6;
+}
+
+message InstanceCreateRequest {
+  string instance = 1;
+  string version = 2;
+  Instance spec = 3;
```
nmiculinic, line 63: Isn't everything you need contained in the Instance? (except account maybe)

pkg/apiserver/api/v1/instance.proto
```diff
@@ -0,0 +1,89 @@
...
+message InstanceGetRequest {
+  string instance = 1;
+  string version = 2;
+  string name = 3;
+  string account = 4;
+}
+
+message InstanceDeleteRequest {
+  string instance = 1;
+  string version = 2;
```
nmiculinic, line 46: What has the version got to do with the deletion request? It shouldn't matter since the version is useful only for get/listing, for deletion is irrelevant.

pkg/apiserver/api/v1/instance.proto
```diff
@@ -0,0 +1,89 @@
...
+  string status = 3;
+}
+
+message InstanceList {
+  ListMeta metadata = 1;
+  repeated Instance items = 2;
+}
+
+message InstanceGetRequest {
+  string instance = 1;
```
nmiculinic, line 38: What's the difference between `instance` and `name`?

aborilov: so, `instance` it's service/offering name, i.e. `couchdbs.eu-west-1.team-a`, `name` that we get from `offering.spec.crd.name`. we can rename this field to `offering` or back to `service`.
and `name` is the name of the instance that created from this offering, like `ProdCouchDB`

pkg/apiserver/api/v1/instance.proto
aborilov: same here
so, `instance` it's service/offering name, i.e. `couchdbs.eu-west-1.team-a`, `name` that we get from `offering.spec.crd.name`. we can rename this field to `offering` or back to `service`.
and `name` is the name of the instance that created from this offering, like `ProdCouchDB`

pkg/apiserver/api/v1/instance.proto
aborilov: to know which instance types do ween need to list, same as `kubectl get pod`. in our case it will be `couchdbs.eu-west-1.team-a`

pkg/apiserver/api/v1/instance.proto
aborilov: UI can get corresponding `offering` using `instance`

pkg/apiserver/api/v1/instance.proto
aborilov: to be consistent with all other methods.
`/v1/accounts/{account}/instance/{instance}/{version}/{name}`
don't want to have empty {version} in that path

pkg/apiserver/api/v1/instance.proto
aborilov: actually, `Account` is also in `Instance.Metadata`, but there are no `version` and `instance` there

pkg/apiserver/api/v1/instance.proto
nmiculinic: But this parameter is unnecessary. Maybe reverse the ordering as:
 
`/v1/accounts/{account}/instance/{instance}/{name}/{version}`

for all other API calls?

At the end of the day, I don't really care that much for REST API, since gRPC-web should be used by default.

pkg/apiserver/api/v1/instance.proto
```diff
@@ -0,0 +1,103 @@
...
+  string version = 2;
+  // Name of the offering instance
+  string name = 3;
+  // Account indicate namespace of the project/account
+  string account = 4;
+}
+
+message InstanceListRequest {
+  // Offering name, i.e. couchdb.eu-west-1.team-a
+  string offering = 1;
```
nmiculinic, line 61: Can I list all instances within an account? Or is this something you'd like handling on the front-end?

pkg/apiserver/api/v1/instance.proto
nmiculinic: Add offering information in the instance.

pkg/apiserver/internal/v1/instance.go
```diff
@@ -0,0 +1,250 @@
...
+		client: c,
+		mapper: mapper,
+	}
+}
+func (o instanceServer) Create(ctx context.Context, req *v1.InstanceCreateRequest) (res *v1.Instance, err error) {
+	obj := &unstructured.Unstructured{}
+
+	gvk, err := o.gvkFromInstance(o.mapper, req.Offering, req.Version)
+	if err != nil {
+		return nil, status.Errorf(codes.Internal, "creating instance: unable to get Kind: %s", err.Error())
```
nmiculinic, line 57: This isn't internal error, but argument error.

pkg/apiserver/api/v1/instance.proto
aborilov: Do you mean all instances of all different offerings? In that case, no, current API can't do that, we can add this later, but it will be separate endpoint(`/v1/accounts/{account}/instances/`

pkg/apiserver/api/v1/instance.proto
aborilov: but for the listing we have `"/v1/accounts/{account}/instances/{offering}/{version}"`, so we can't revers that

pkg/apiserver/api/v1/instance.proto
nmiculinic: ok

pkg/apiserver/internal/v1/instance.go
aborilov: done

aborilov:
> Please add integration test for the services, under `/test/integration/apiserver.go`

done

pkg/apiserver/api/v1/instance.proto
aborilov: done

pkg/apiserver/api/v1/instance.proto
```diff
@@ -0,0 +1,104 @@
...
+
+import "google/api/annotations.proto";
+import "google/protobuf/empty.proto";
+
+import "meta.proto";
+
+message Instance {
+  ObjectMeta metadata = 1;
+  string offering = 2;
+  string spec = 3;
```
thetechnick, line 29: Could you please add a wrapper object for arbitrary data? (`status` and `spec` fields)
I would like to have the possibility to support multiple encoding styles the future.
e.g. `JSON` or `YAML` - maybe even nested protobuf

```
message RawObject {
  string encoding = 1;
  bytes data = 2;
}
```
Please use `JSON` as default for know, although I already see a use-case for the UI requesting the nested object as `YAML` (When editing the YAML directly)

pkg/apiserver/api/v1/instance.proto
```diff
@@ -0,0 +1,104 @@
...
+option go_package = "v1";
+
+import "google/api/annotations.proto";
+import "google/protobuf/empty.proto";
+
+import "meta.proto";
+
+message Instance {
+  ObjectMeta metadata = 1;
+  string offering = 2;
```
thetechnick, line 28: Ok for this PR, but I see us changing this into a common `kind`/`apiVersion` field similar to the Kubernetes one.
Although we need to specify the mapping of Kubernetes apiVersion to "Our apiVersion", especially for these dynamic services.

pkg/apiserver/api/v1/instance.proto
```diff
@@ -0,0 +1,115 @@
...
+  // Status - instance status
+  RawObject status = 4;
+}
+
+message InstanceList {
+  ListMeta metadata = 1;
+  repeated Instance items = 2;
+}
+
+message InstanceGetRequest {
```
jiachengxu, line 42: Since PR #426 is merged, can you please reuse/extend the Get/ListRequest in Service API?

aborilov: but Service API has additional params, that no present in `Get/ListRequest`

pkg/apiserver/api/v1/instance.proto
aborilov: done

pkg/apiserver/api/v1/instance.proto
jiachengxu: I think you can extend Get/ListRequest with new params. I also use Get/ListRequest in the Account PR.

aborilov: yes, because account service doesn't have any additional fields

jiachengxu: OK, so what is the reason to have a general List/GetRequest? I don't see any reason that we can not extend it to include params and use it everywhere

pkg/apiserver/api/v1/util.go
```diff
@@ -0,0 +1,58 @@
...
+	"encoding/json"
+	"errors"
+	fmt "fmt"
+
+	"sigs.k8s.io/yaml"
+)
+
+const (
+	FormatJSON = "json"
+	FormatYAML = "yaml"
```
thetechnick, line 29:
```suggestion
	EncodingJSON = "json"
	EncodingYAML = "yaml"
```

Just a nit :)

aborilov: done

test/integration/servicecluster.go
```diff
@@ -123,7 +123,7 @@ func newServiceClusterSuite(
 			},
 		}
 		require.NoError(t, managementClient.Create(ctx, catalogEntrySet))
-		require.NoError(t, testutil.WaitUntilReady(ctx, managementClient, catalogEntrySet, testutil.WithTimeout(2*time.Minute)))
+		require.NoError(t, testutil.WaitUntilReady(ctx, managementClient, catalogEntrySet, testutil.WithTimeout(time.Minute)))
```
jiachengxu, line 126: Why do you change this timeout back?

pkg/testutil/framework.go
```diff
@@ -232,6 +232,60 @@ func (f *Framework) NewFakeDB(name, namespace string) *fakev1alpha1.DB {
 	}
 }
 
+func (f *Framework) NewServiceCluster(name, namespace, secret string) *corev1alpha1.ServiceCluster {
```
jiachengxu, line 235: How about we remove the Framework pointer receiver of these helper functions and add them in https://github.com/kubermatic/kubecarrier/blob/offering-watch-api/pkg/testutil/helpers.go, the reason is because:
1. the framework receiver is unused in these helper functions, and doesn't make sense to have it.
2. Then you don't need to change the of the following offeringService, regionService, providerService in the following just for using helper functions.
3. Our unit tests can also use them in the future.

test/integration/apiserver.go
```diff
@@ -197,22 +197,170 @@ func newAPIServer(f *testutil.Framework) func(t *testing.T) {
...
 	}
 }
 
+func instanceService(ctx context.Context, conn *grpc.ClientConn, managementClient *testutil.RecordingClient, f *testutil.Framework) func(t *testing.T) {
+	return func(t *testing.T) {
+		serviceClient, err := f.ServiceClient(t)
+		require.NoError(t, err, "creating service client")
+		ctx, cancel := context.WithCancel(context.Background())
+		t.Cleanup(cancel)
+		managementClient, err := f.ManagementClient(t)
```
jiachengxu, line 223: I don't think you need the managementClient and create a new context in the function, you can just use them by passing from the function, right?

test/integration/servicecluster.go
aborilov: because I thought that the problem was with `catalogentryset` cleanup, but apparently it also needs to be increased here

test/integration/apiserver.go
aborilov: you are right

pkg/testutil/framework.go
aborilov: then I will move all `NewSomething` helpers

jiachengxu: Yeah, would be really great, I have moved the NewProviderAccount and stuff in my account PR also.

jiachengxu: You can move helpers that are used by your PR, and then we move the rest of helpers gradually :)

aborilov: too late, already moved everything :)

aborilov: but I still need `func instanceService(ctx context.Context, conn *grpc.ClientConn, managementClient *testutil.RecordingClient, f *testutil.Framework) func(t *testing.T) {
` in my tests, so have to change the signatures of the following offeringService, regionService, providerService

test/integration/apiserver.go
aborilov: done

pkg/testutil/framework.go
aborilov: done

PR #426, opened 2020-05-18
[API Server] Use general ListRequest and GetRequest in services

jiachengxu:
Currently, for our services, for example, offering service, region service, they have dedicated ListRequest and GetRequest protobuf definitions, this PR unifies them into general ListRequest and GetRequest.

Fixes #

**Documentation**:

```release-note
NONE
```


pkg/apiserver/internal/util/util.go
```diff
@@ -28,3 +34,38 @@ func TimestampProto(t *metav1.Time) (*timestamp.Timestamp, error) {
 	}
 	return ptypes.TimestampProto(t.Time)
 }
+
+func ValidateGetRequest(req *v1.GetRequest) error {
```
aborilov, line 38: I thought we want validation to be versioned.
maybe `pkg/apiserver/internal/v1/validation.go` is a better place for that, what do you think?

jiachengxu: Yeah, agree, that makes more sense, will do that

jiachengxu: Done

aborilov:
@jiachengxu overall looks good, only one question, what do you think about renaming request to `GetAccountRequest` and `ListAccountRequest`, or `AccountGet/ListRequest`.
I want to show somehow that those requests are account scoped because we will have some other not account scoped request, lile for `Account` service that you already added and may be some other for admin interface

jiachengxu:
> @jiachengxu overall looks good, only one question, what do you think about renaming request to `GetAccountRequest` and `ListAccountRequest`, or `AccountGet/ListRequest`.
> I want to show somehow that those requests are account scoped because we will have some other not account scoped request, lile for `Account` service that you already added and may be some other for admin interface

If I recall correctly, we are going to use the single GetReqest/ListRequest for all our services(not only account scoped and also other services, for example account service)? and we just implement different validateRequest method and ignore query parameters accordingly?

If you want to rename to show that Request is account scoped, we need to be careful :) because I have already called the ListRequest to list account in Account service to `ListAccountRequest` :)

aborilov:
> because I have already called the ListRequest to list account in Account service to `ListAccountRequest` :)
yeah, I saw that.
>If I recall correctly, we are going to use the single GetReqest/ListRequest for all our services(not only account scoped and also other services, for example account service)? and we just implement different validateRequest method and ignore query parameters accordingly?
and I remember that discussion, I that case we need to rename `validateGetRequest` :) to show that it will validate `account` also

PR #437, opened 2020-05-26
[API Server] Account Service

jiachengxu:

Fixes #

**Documentation**:

```release-note
NONE
```


pkg/apiserver/api/v1/account.proto
```diff
@@ -0,0 +1,84 @@
...
+
+message Account {
+  ObjectMeta metadata = 1;
+  AccountSpec spec = 2;
+  AccountStatus status = 3;
+}
+
+message AccountSpec {
+  AccountMetadata metadata = 1;
+  repeated string roles = 2;
```
nmiculinic, line 35: Can this be an enum instead of string?

pkg/apiserver/api/v1/account.proto
```diff
@@ -0,0 +1,84 @@
...
+}
+
+message Subject {
+  string kind = 1;
+  string apiGroup = 2;
+  string name = 3;
+  string namespace = 4;
+}
+
+message AccountStatus {
```
nmiculinic, line 54: Is this too much detail exposed in the API server? Why would the end-user care which namespace the account is in? Or the accounts `observedGeneration`?

pkg/apiserver/api/v1/account.proto
```diff
@@ -0,0 +1,84 @@
...
+  ObjectReference namespace = 1;
+  int64 observedGeneration = 2;
+  string phase = 3;
+  repeated AccountCondition conditions = 4;
+}
+
+message AccountCondition {
+  string type = 1;
+  string status = 2;
```
nmiculinic, line 63: can this be an enum?

pkg/apiserver/apiserver.go
```diff
@@ -127,6 +128,35 @@ func runE(flags *flags, log logr.Logger) error {
...
+		return fmt.Errorf("fail to register field index for Account Username: %w", err)
+	}
+	accountClient := &client.DelegatingClient{
+		Reader:       accountCache,
+		Writer:       c,
+		StatusClient: c,
+	}
+	log.Info("start cache")
+	go func() {
+		if err := accountCache.Start(ctrl.SetupSignalHandler()); err != nil {
```
nmiculinic, line 149: Let's have main scoped context, and not create new one at each turn. (Replacing both `context.Background()` as well as `ctrl.SetupSignalHandler()` )

pkg/testutil/helpers.go
```diff
@@ -287,3 +288,46 @@ func componentCheck(
 	}
 	t.Logf(fmt.Sprintf("%s is ready", componentName))
 }
+
+func NewProviderAccount(name string, subjects ...rbacv1.Subject) *catalogv1alpha1.Account {
```
nmiculinic, line 292: This should be in a separate PR in the future since refactoring `framework.go` is orthogonal to the account service. You could make stacked PR, and we'd quickly approve the refactoring PR.

pkg/apiserver/internal/v1/account.go
```diff
@@ -0,0 +1,174 @@
...
+// +kubebuilder:rbac:groups=catalog.kubecarrier.io,resources=accounts,verbs=get;list;watch
+
+func NewAccountServiceServer(c client.Client) v1.AccountServiceServer {
+	return &accountServer{
+		client: c,
+	}
+}
+
+func (o accountServer) List(ctx context.Context, req *v1.AccountListRequest) (res *v1.AccountList, err error) {
+	user, present := oidc.ExtractUserInfo(ctx)
```
nmiculinic, line 49: As a side note, maybe when the user isn't authenticated, we could inject the `system:anonymous` user by default. 

Right now all endpoints are under OIDC middleware, and must pass auth procedure. This would allow more granular permissions, for example, the version endpoint might as well be available publically. 

https://kubernetes.io/docs/reference/access-authn-authz/authentication/#anonymous-requests

pkg/apiserver/internal/v1/account.go
```diff
@@ -0,0 +1,174 @@
...
+}
+
+func (o accountServer) handleListRequest(ctx context.Context, req *v1.AccountListRequest, username string) (res *v1.AccountList, err error) {
+	var listOptions []client.ListOption
+	listOptions, err = o.validateListRequest(req)
+	listOptions = append(listOptions, accountByUsernameListOption(username))
+	if err != nil {
+		return nil, status.Error(codes.InvalidArgument, err.Error())
+	}
+	accountList := &catalogv1alpha1.AccountList{}
```
nmiculinic, line 63: In the future, we should filter the account list depending on which accounts the users has meaningful permissions towards.

pkg/apiserver/api/v1/account.proto
jiachengxu: There is some discussion around enum/string in this PR and also #421 . I am not a big fan of enum in protobuf, since it is really hard to extend in the future, it may cause issues if you add more values to it in the future, the different versions between client and server can lead to issues which are hard to debug.
Also, if you check the kube-apiserver, they also don't use enum and use strings.

Now I am thinking about if I should put the string to a dedicated message, for example:
```
message Role {
string type =1; 
}
```
In this way, at least it is typed I think? WDYT?

pkg/apiserver/api/v1/account.proto
jiachengxu: For namespace, user cares, because user should know where all his/her objects/ service instances will be created.
For `observedGeneration`, WDYT @thetechnick @aborilov ?

pkg/testutil/helpers.go
jiachengxu: This should be in this PR, because the account integration test(which doesn't have a framwork pointer) needs to use it, that's why I remove the receiver.

pkg/apiserver/internal/v1/account.go
jiachengxu: Yeah, that's a good catch, but do you actually want anonymous users to list accounts?

pkg/apiserver/internal/v1/account.go
jiachengxu: What do you mean by users have meaningful permissions towards?
Currently, we are listing account which has the user in its `.Spec.Subjects` which means we are listing account which user has access to.

pkg/apiserver/api/v1/account.proto
nmiculinic: > since it is really hard to extend in the future

what makes you think it's hard to extend in the future?

https://medium.com/@akhaku/protobuf-definition-best-practices-87f281576f31 (under 6) is a good suggestion for how to handle the unknown enum fields. 

And yeah, my biggest issue with naked meaningful strings is that they are both untyped and their allowed values aren't documented within the .proto file. What I like about enums is they're both typed and enumerate all possible valid states.

In this case, either enum or leave it as a string, creating an additional type looks like overkill for now. 

The biggest enum drawback is if the client encounters `UNKNOWN_ROLE` it cannot display it to the user in a meaningful way, aside from that I don't see issues with it.

pkg/apiserver/internal/v1/account.go
nmiculinic: Nope; my thought process is you shouldn't handle auth explicitly in every API call. ( `if !present ..`). It should fail naturally when you try using the client with the `system:anonymous` user.

It seems kinda silly both having an authentication middleware and you still have to check in every endpoint explicitly is the user authenticated.

pkg/apiserver/internal/v1/account.go
nmiculinic: oh ok; I haven't noticed you're doing filtering via the field index; this is fine.

pkg/apiserver/api/v1/account.proto
thetechnick: Kubernetes decided against Enums in their APIs - and I would stay with that for our API:
- https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md
- https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api_changes.md
(Search for `enum`)

pkg/apiserver/api/v1/account.proto
thetechnick: Users don't need to know the namespace, as we don't expose access to "Namespaces" via the KubeCarrier API.
While not critical, please remove the field.

pkg/apiserver/internal/v1/account.go
thetechnick: This looks like we should put this into a grpc middleware and just extract the authenticated userinfo from the context for authorization.

Maybe we can use:
https://github.com/grpc-ecosystem/go-grpc-middleware
(This adds even more capability for tracing and monitoring)

Seems like you can even specify advanced validation:
https://github.com/grpc-ecosystem/go-grpc-middleware/blob/master/validator/doc.go

Issues: https://github.com/kubermatic/kubecarrier/issues/441

pkg/apiserver/api/v1/account.proto
jiachengxu: fixed

pkg/apiserver/apiserver.go

pkg/apiserver/api/v1/account.proto
nmiculinic: ok, this doesn't change the fact we have enum like strings in our API. Could we at least document in the `.proto` files their expected values?

pkg/apiserver/api/v1/account.proto
nmiculinic: Why are we exposing `phase` here which is deprecated in the Kubernetes?

pkg/apiserver/api/v1/account.proto
```diff
@@ -0,0 +1,91 @@
...
+
+message AccountCondition {
+  AccountConditionType type = 1;
+  ConditionStatus status = 2;
+  google.protobuf.Timestamp lastTransitionTime = 3;
+  string reason = 4;
+  string message = 5;
+}
+
+message ConditionStatus {
```
nmiculinic, line 72: This could be moved to `types.proto` or something like that; no need to repeat it everywhere.

pkg/apiserver/apiserver.go
```diff
@@ -127,23 +128,53 @@ func runE(flags *flags, log logr.Logger) error {
...
+	}
+	if err := v1.RegisterAccountUsernameFieldIndex(accountCache); err != nil {
+		return fmt.Errorf("fail to register field index for Account Username: %w", err)
+	}
+	accountClient := &client.DelegatingClient{
+		Reader:       accountCache,
+		Writer:       c,
+		StatusClient: c,
+	}
+	ctx := context.Background()
```
nmiculinic, line 147: You could make this cancelable context, and kill it when the `ctrl.SetupSignalHandlers()` terminate.

jiachengxu: What is the benefit of this approach? Why don't we just pass the same ctx also in `cache.Start`?

pkg/apiserver/api/v1/account.proto
jiachengxu: fixed

pkg/apiserver/api/v1/account.proto
jiachengxu: hmm, you are right, removed

pkg/apiserver/apiserver.go
jiachengxu: I see your point, fixed

pkg/apiserver/internal/v1/account.go
```diff
@@ -0,0 +1,172 @@
...
+
+// +kubebuilder:rbac:groups=catalog.kubecarrier.io,resources=accounts,verbs=get;list;watch
+
+func NewAccountServiceServer(c client.Client) v1.AccountServiceServer {
+	return &accountServer{
+		client: c,
+	}
+}
+
+func (o accountServer) List(ctx context.Context, req *v1.AccountListRequest) (res *v1.AccountList, err error) {
```
aborilov, line 48: we can't test this method as for now, because we can ExtractUser from context, one of the solutions can be to add use dependency injection and pass `oidc` as a dependency to `NewAccountServiceServer` and then we can mock it in the tests

jiachengxu: We are currently unit-testing the `handleListRequest`, I think it should be enough, for List functionality, we are testing it in e2e tests

aborilov: ok for this PR, but if we will have more and more of such places, we need to think about solution

pkg/apiserver/internal/v1/account_test.go
```diff
@@ -0,0 +1,238 @@
...
+							},
+							Subjects: []*v1.Subject{
+								{
+									Kind:     "Group",
+									ApiGroup: "rbac.authorization.k8s.io",
+									Name:     "provider1",
+								},
+							},
+						},
+						Status: &v1.AccountStatus{},
```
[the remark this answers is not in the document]
jiachengxu: done

pkg/apiserver/apiserver.go
nmiculinic: I had something else in mind; something like this: 

```
go func() {
   <-ctrl.SetupSignalHandler()
   cancel()
}
```

and using the `ctx.Done()` when someone requests the stop chan.

pkg/apiserver/internal/v1/account.go
nmiculinic: This is for further improvement right? Not blocking this PR merging

PR #459, opened 2020-06-08
Refactored authN handling in the API server

nmiculinic:
**What this PR does / why we need it**:

Refactored authentication handling to plugin-based approach for better extensibility. This is a prerequisite for:

* #445 
* #447 

```release-note
NONE
```

PR #460, opened 2020-06-08
KubeCarrier deploys API Server

### LINKED ISSUE #446: API Server Deployment
The KubeCarrier API server should be present by default in every KubeCarrier installation, although custom configuration might be needed to control ingress traffic to the API server and it's serving certificates.

**Example Configuration**

```yaml
apiVersion: operator.kubecarrier.io/v1alpha1
kind: KubeCarrier
metadata:
  name: kubecarrier
spec:
  api:
    # Name of a secret within the kubecarrier-system namespace
    #
    # optional
    # if not present a self-signed certificate for localhost and
    # 127.0.0.1 will be generated via cert-manager
    tlsSecretRef:
      name: kubecarrier-tls

    # OIDC configuration
    #
    # optional
    # if not present, OIDC authentication is disabled and only ServiceAccount tokens auth is supported.
    # see issue #445
    oidc: {}
```

**Acceptance Criteria**
- KubeCarrier API server is deployed when a `KubeCarrier` instance is created
- advanced configuration of the API server is optional

jiachengxu:

Fixes #446

**Documentation**:

```release-note
NONE
```


pkg/apis/operator/v1alpha1/apiserver_types.go
```diff
@@ -18,84 +18,11 @@ package v1alpha1
...
-	//
-	// https://openid.net/specs/openid-connect-core-1_0.html#IDTokenValidation
-	// +kubebuilder:default=RS256;
-	SupportedSigningAlgs []string `json:"supportedSigningAlgs,omitempty"`
-
-	// RequiredClaims, if specified, causes the OIDCAuthenticator to verify that all the
-	// required claims key value pairs are present in the ID Token.
-	// +optional
-	RequiredClaims map[string]string `json:"requiredClaims,omitempty"`
+	API APIServerConfig `json:"api,omitempty"`
```
thetechnick, line 25: This does not need to be nested, please leave these fields in `APIServerSpec` and reference the `APIServerSpec` as `api` key from the `KubeCarrierSpec`.

e.g.

```golang
type KubeCarrierSpec struct {
	// +optional
	API APIServerSpec `json:"api,omitempty"`
}
```

I would like to avoid calling objects `*Config` - as all these APIs are declarative, they are all "Configs" or Specifications and `Spec` is way more common in k8s objects.

jiachengxu: Yeah, make sense, will do that

pkg/apiserver/apiserver.go
```diff
@@ -85,7 +86,10 @@ func NewAPIServer() *cobra.Command {
 	cmd.Flags().StringVar(&flags.TLSCertFile, "tls-cert-file", "", "File containing the default x509 Certificate for HTTPS. If not provided no TLS security shall be enabled")
 	cmd.Flags().StringVar(&flags.TLSPrivateKeyFile, "tls-private-key-file", "", "File containing the default x509 private key matching --tls-cert-file.")
 	cmd.Flags().StringArrayVar(&flags.CORSAllowedOrigins, "cors-allowed-origins", []string{"*"}, "List of allowed origins for CORS, comma separated. An allowed origin can be a regular expression to support subdomain matching. If this list is empty CORS will not be enabled.")
-	oidc.AddOIDCPFlags(&flags.OIDCOptions, cmd.Flags())
+	cmd.Flags().BoolVar(&flags.enableOIDC, "enable-oidc", false, "List of allowed origins for CORS, comma separated. An allowed origin can be a regular expression to support subdomain matching. If this list is empty CORS will not be enabled.")
```
thetechnick, line 89: What about we define this as a list of auth-providers?

e.g. `--authentication-mode=OIDC,Static,ServiceAccount`

ref from Kubernetes:
```
--authorization-mode stringSlice     Default: [AlwaysAllow]
--
  | Ordered list of plug-ins to do authorization on secure port. Comma-delimited list of: AlwaysAllow,AlwaysDeny,ABAC,Webhook,RBAC,Node.
```

jiachengxu: Yeah, I put it as a simple boolean because Neven has already done that in #459, and I don't want to repeat that PR here, after that PR is merged, will also adapt the changes here.

pkg/apis/operator/v1alpha1/apiserver_types.go
jiachengxu: done

pkg/apiserver/apiserver.go

pkg/testutil/framework.go
```diff
@@ -25,7 +25,7 @@ import (
 	"testing"
 	"time"
 
-	certmanagerv1alpha3 "github.com/jetstack/cert-manager/pkg/apis/certmanager/v1alpha3"
+	certmanagerv1alpha2 "github.com/jetstack/cert-manager/pkg/apis/certmanager/v1alpha2"
```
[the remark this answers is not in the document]
jiachengxu: We are using v1alpha2 across our projects, including kustomize configurations and reconcile packages, I would use the same version everywhere, if you want to upgrade it, we should upgrade everywhere.

pkg/internal/resources/apiserver/apiserver.go
```diff
@@ -48,133 +48,194 @@ var k = kustomize.NewDefaultKustomize()
...
-								},
-								{
-									"name":  "API_SERVER_TLS_PRIVATE_KEY_FILE",
-									"value": "/run/serving-certs/tls.key",
-								},
-								{
-									"name":  "API_SERVER_OIDC_ISSUER_URL",
-									"value": c.Spec.OIDC.IssuerURL,
+	var managerEnv map[string]interface{}
+	if c.Spec.OIDC != nil {
```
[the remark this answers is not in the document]
jiachengxu: I also wanted to use `Deployment` to deduplicate this, just remembered that we had issues with it(some fields that we did't want to). @thetechnick Do you remember what was the issue?

pkg/internal/resources/apiserver/apiserver.go
```diff
@@ -48,133 +48,194 @@ var k = kustomize.NewDefaultKustomize()
...
+				"name":      "manager",
+				"namespace": "system",
+			},
+			"spec": map[string]interface{}{
+				"template": map[string]interface{}{
+					"spec": map[string]interface{}{
+						"containers": []map[string]interface{}{
+							{
+								"name": "manager",
+								"args": append([]string{
```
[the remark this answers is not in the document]
jiachengxu: done

pkg/internal/resources/apiserver/apiserver.go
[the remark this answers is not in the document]
jiachengxu: I have tried to swtich to using `Deployment`, and I looked into our golden file, only `deployment.Spec.Strategy` and `deployment.Status` are filled with `{}` empty values, so I think it should be fine? and if we still want to remove these empty values, we can still convert `Deployment` to `Unstructured` and remove these keys before marshaling.

Also, I would like to use `Deployment` rather than `map[string]interface{}`, because something can go wrong easily with map, for example, the key in our `VolumeMounts` https://github.com/kubermatic/kubecarrier/blob/feature-api-server/pkg/internal/resources/apiserver/apiserver.go#L155
It should be `readOnly` not `readyOnly`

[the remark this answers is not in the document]
nmiculinic: yes yes yes! on not using `map[string]interface{}` 

hmm, which serializer/encoded have you used? for `.status` I'm not at all worried about since this wouldn't be applied to the objects due to `/status` subresource.

for `.spec.strategy` would the Kubernetes API correctly apply the defaults here? If so...than let's roll with it!

pkg/testutil/framework.go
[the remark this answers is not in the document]
nmiculinic: Sure; let's have whole project transition to `v1alpha3` version. Should be more or less straightforward. Could you do it in some subsequent PRs?

pkg/apiserver/apiserver.go
```diff
@@ -97,7 +97,7 @@ func NewAPIServer() *cobra.Command {
 	cmd.Flags().StringVar(&flags.TLSCertFile, "tls-cert-file", "", "File containing the default x509 Certificate for HTTPS. If not provided no TLS security shall be enabled")
 	cmd.Flags().StringVar(&flags.TLSPrivateKeyFile, "tls-private-key-file", "", "File containing the default x509 private key matching --tls-cert-file.")
 	cmd.Flags().StringArrayVar(&flags.CORSAllowedOrigins, "cors-allowed-origins", []string{"*"}, "List of allowed origins for CORS, comma separated. An allowed origin can be a regular expression to support subdomain matching. If this list is empty CORS will not be enabled.")
```
nmiculinic, line 99: could you also move this to `StringSliceVar`?

pkg/apis/operator/v1alpha1/apiserver_types.go
```diff
@@ -24,9 +24,11 @@ import (
 // APIServerSpec defines the desired state of APIServer
 type APIServerSpec struct {
 	// TLSSecretRef references the TLS certificate and private key for serving the KubeCarrier API.
-	TLSSecretRef ObjectReference `json:"tlsSecretRef"`
+	// +optional
+	TLSSecretRef *ObjectReference `json:"tlsSecretRef,omitempty"`
 	// OIDC specifies OpenID Connect configuration for API Server authentication
-	OIDC APIServerOIDCConfig `json:"oidc"`
+	// +optional
+	OIDC *APIServerOIDCConfig `json:"oidc,omitempty"`
```
nmiculinic, line 31: Could you add here `AnonymousAuthConfig` pointer field?  I'd love allowing user to include the Anonymous auth if they wish to, orthogonal whether the OIDC is included or not.  Thus something like this:

```yaml
    anonymousAuth: 
        enabled: true
```

Possible extension in  the future:

```yaml
    anonymousAuth: 
        enabled: true
        user: ...
        group: [] 
```

Thus in the resource creation, you check for all auth providers with non-nil config and include them in the API server setup. The ordering should be "OIDC,Anonymous";  and maybe in the future, we could allow users to set the ordering explicitly. 

At the end of the day, the user should be empowered picking and choose which auth providers he's going to use and whether he wants anonymous auth enabled or not. 

After this is implemented I'll LGTM it.

P.S.
alternative YAML schema:

```
authentication:
- oidc: {...}
- anonymous: {enabled: true} 
```

and this would also configure auth provider ordering explicitly by the user, and not our own heuristics. 

@thetechnick

pkg/apiserver/apiserver.go
thetechnick: As the input pointer is `<nil>`, I guess the flag will not work atm - right?

pkg/apis/operator/v1alpha1/apiserver_types.go
thetechnick: I like the idea, but I would like to see this done in another PR.
(this one is already fairly large, due to all the boilerplating)

nmiculinic: https://github.com/kubermatic/kubecarrier/issues/470

pkg/apiserver/apiserver.go

pkg/internal/resources/apiserver/apiserver.go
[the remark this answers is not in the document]
jiachengxu: actually I am just using `yaml.Marshal()` and it works :)

PR #479, opened 2020-06-16
Watch test refactor

nmiculinic:

This is attempt into making watch e2e test more stable

```release-note
NONE
```

### DIFF
--- a/test/integration/apiserver.go
+++ b/test/integration/apiserver.go
@@ -789,26 +789,23 @@ func offeringService(ctx context.Context, conn *grpc.ClientConn, account *catalo
 			LabelSelector: "test-label==offering1",
 		})
 		require.NoError(t, err)
+		nextEventType(t, watchClient, watch.Added)
+
 		// Update an offering object to get Modified event.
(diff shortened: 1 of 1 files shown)


test/integration/apiserver.go
```diff
@@ -789,28 +789,25 @@ func offeringService(ctx context.Context, conn *grpc.ClientConn, account *catalo
 			LabelSelector: "test-label==offering1",
 		})
 		require.NoError(t, err)
+		nextEventType(t, watchClient, watch.Added)
+
 		// Update an offering object to get Modified event.
 		offering1.Spec.Metadata.ShortDescription = "test offering update"
 		require.NoError(t, managementClient.Update(ctx, offering1))
+		nextEventType(t, watchClient, watch.Modified)
```
jiachengxu, line 797: I am not sure if this will help, because sometimes the `Modified` event will be lost, @aborilov and I have seen this quite a lot

nmiculinic: They could be lost due to delay between create and delete?  

Otherwise we have to drop this test...

PR #490, opened 2020-06-22
add util watch method

aborilov:

Fixes #

**Documentation**:

```release-note
NONE
```

### DIFF
--- a/pkg/apiserver/internal/v1/util.go
+++ b/pkg/apiserver/internal/v1/util.go
@@ -17,11 +17,18 @@ limitations under the License.
 package v1
 
 import (
+	"context"
+
 	"github.com/golang/protobuf/ptypes"
+	"github.com/golang/protobuf/ptypes/any"
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/status"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/runtime"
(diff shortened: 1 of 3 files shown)


pkg/apiserver/internal/v1/util.go
```diff
@@ -158,3 +165,35 @@ func convertListMeta(in metav1.ListMeta) (out *v1.ListMeta) {
 	}
 	return
 }
+
+type ConvertFunc func(runtime.Object) (*any.Any, error)
+type SendFunc func(*v1.WatchEvent) error
+
+func watch(ctx context.Context, client dynamic.Interface, gvr schema.GroupVersionResource, namespace string, opts metav1.ListOptions, sendFunc SendFunc, convertFunc ConvertFunc) error {
```
jiachengxu, line 172: why do we need to pass `ctx` and `sendFunc` to the function? Why don't we just pass `stream` to the function?

aborilov: because each server has own stream, with a different type
https://github.com/kubermatic/kubecarrier/blob/ace10a9bd4cfd2287a523dc9b588371c50270574/pkg/apiserver/internal/v1/offering.go#L125.
we can add create an interface with `Context` and `Send` methods and pass `stream` as it implements it, but I think it will add complexity.

jiachengxu: For watch function working properly, the context should always be the `Stream.Context()` right? If someone pass a random context to it, it will not work I guess, so I think it is better to pass the stream to it? WDYT?

aborilov: I made this change, pls take a look.
I also like it, fewer params to pass

PR #497, opened 2020-06-29
add `paused` flag to catapult

### LINKED ISSUE #341: Catapult paused flag

aborilov:
Allow pausing reconciliation of catapult managed instances.

Fixes #341 

**Documentation**:

```release-note
NONE
```


pkg/apis/operator/v1alpha1/catapult_types.go
```diff
@@ -36,6 +36,8 @@ type CatapultSpec struct {
 	// ServiceCluster: Webhook will call webhooks of the CRD in the ServiceCluster with dry-run flag.
 	// +kubebuilder:default:=None
 	WebhookStrategy corev1alpha1.WebhookStrategyType `json:"webhookStrategy,omitempty"`
+	// Paused tell controller to pause reconciliation process and assume that Catapult is ready
+	Paused bool `json:"paused,omitempty"`
```
jiachengxu, line 40: How about we use `string` type `PausedCondition`: (`True`, `False`) for this? I think boolean variable is not the preferable way in Kubernetes since it is hard to extand.

pkg/operator/internal/controllers/catapult_controller.go
```diff
@@ -73,6 +73,15 @@ func (r *CatapultReconciler) Reconcile(req ctrl.Request) (ctrl.Result, error) {
...
 		return ctrl.Result{}, client.IgnoreNotFound(err)
 	}
+	if catapult.Spec.Paused {
+		if catapult.SetPausedCondition() {
+			if err := r.Client.Status().Update(ctx, catapult); err != nil {
+				return ctrl.Result{}, fmt.Errorf("updating %s status: %w", catapult.Name, err)
+			}
+			// reconciliation paused, skip all other handlers
+			return ctrl.Result{}, nil
+		}
```
jiachengxu, line 83: What if catapult is already paused, and `SetOausedCondition` returns `false`? 
It should also skip reconciliation right?

pkg/apis/operator/v1alpha1/catapult_types.go
aborilov: yeah, good point, always forget about that

pkg/operator/internal/controllers/catapult_controller.go
aborilov: oh, good catch, this return should be in the outer block

aborilov: done

pkg/apis/operator/v1alpha1/catapult_types.go
aborilov: done

pkg/apis/operator/v1alpha1/catapult_types.go
```diff
@@ -36,8 +36,23 @@ type CatapultSpec struct {
...
+type PausedFlagType string
+
+func (o PausedFlagType) IsPaused() bool {
+	return o == PausedFlagTrue
+}
+
+// Values of PausedFlagType.
+const (
+	PausedFlagTrue  PausedFlagType = "true"
+	PausedFlagFalse PausedFlagType = "false"
```
jiachengxu, line 53: Can you please make these two string values to `True` and `False` Here is the kubernetes api-conventions:https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#constants

aborilov: Done

PR #499, opened 2020-07-02
add auth schema with auth providers order

### LINKED ISSUE #470: API Server CRD auth schema
https://github.com/kubermatic/kubecarrier/pull/460#discussion_r439323997

As continued from this comment here's my proposed API Server CRD schema for authentication:

```yaml
authentication:
- oidc: {...}
- anonymous: {enabled: true} 
```

the list ordering would carry over in the authentication-mode ordering in the API server CLI flag, and finally how the authentication providers shall be evaluated.

aborilov:
configure auth provider ordering explicitly by the user

Fixes #470 

**Documentation**:

```release-note
NONE
```


pkg/apis/operator/v1alpha1/apiserver_types.go
```diff
@@ -26,12 +26,21 @@ type APIServerSpec struct {
 	// TLSSecretRef references the TLS certificate and private key for serving the KubeCarrier API.
 	// +optional
 	TLSSecretRef *ObjectReference `json:"tlsSecretRef,omitempty"`
+	// +optional
+	// Authentication configuration
+	Authentication *Authentication `json:"authentication,omitempty"`
```
jiachengxu, line 31: This should not be optional I think? Otherwise what authentication mode will be enabled if it is omitted? (Anonymous should be enabled as default I think)

pkg/apis/operator/v1alpha1/apiserver_types.go
```diff
@@ -26,12 +26,21 @@ type APIServerSpec struct {
...
+type Authentication struct {
 	// OIDC specifies OpenID Connect configuration for API Server authentication
 	// +optional
 	OIDC *APIServerOIDCConfig `json:"oidc,omitempty"`
 	// StaticUsers specifies static users configuration for API Server authentication
 	// +optional
 	StaticUsers *StaticUsers `json:"staticUsers,omitempty"`
+	// Order in which auth providers executed. Available values: Anonymous, Htpasswd, OIDC, Token
+	// +optional
+	Order []string `json:"order,omitempty"`
```
jiachengxu, line 43: If we are going to use `Order` list, we need to add a webhook to check:
1. `OIDC` != nil if OIDC is in `Order` list
2. `StaticUser` != nil if Htpasswd is in `Order` list

jiachengxu: Also, I think the `Order` list should not be `omitempty`? what will happen in your case if the `Order` is empty?

aborilov: > Also, I think the `Order` list should not be `omitempty`? what will happen in your case if the `Order` is empty?

I will add `Anonymous` by default

> If we are going to use `Order` list, we need to add a webhook to check:
> 
> 1. `OIDC` != nil if OIDC is in `Order` list
> 2. `StaticUser` != nil if Htpasswd is in `Order` list

we don't have webhooks for apiserver as for now, and I don't think it worth adding it for that, what we can easily do, if config is `nil` I can delete this provider from `Order`

jiachengxu: No, we only need to add logic to operator webhook: https://github.com/kubermatic/kubecarrier/blob/master/pkg/operator/internal/webhooks/kubecarrier_webhook.go, since apiserver.spec is enabled in kubecarrier

pkg/apis/operator/v1alpha1/apiserver_types.go
aborilov: I current implementation we have everything optional and we are adding `Anonymous` and `Token` providers by default.
I can add `Anonymous` provider to auth modes in `Order` is empty here
https://github.com/kubermatic/kubecarrier/blob/7c27999632942978b0f471738d8c6760098000e7/pkg/internal/resources/apiserver/apiserver.go#L223
what do you think?

pkg/apis/operator/v1alpha1/apiserver_types.go
aborilov: oh, ok, will do that

pkg/apis/operator/v1alpha1/apiserver_types.go
nmiculinic: > This should not be optional I think? Otherwise what authentication mode will be enabled if it is omitted? (Anonymous should be enabled as default I think)

If the things is optional the defaulting webhook should fill the defaults in.

pkg/operator/internal/webhooks/kubecarrier_webhook.go
```diff
@@ -70,5 +71,33 @@ func (r *KubeCarrierWebhookHandler) validateCreate(kubeCarrier *operatorv1alpha1
 	if kubeCarrier.Name != constants.KubeCarrierDefaultName {
 		return fmt.Errorf("KubeCarrier object name should be 'kubecarrier', found: %s", kubeCarrier.Name)
 	}
+	auth := map[string]bool{}
+	for _, a := range kubeCarrier.Spec.API.Authentication {
+		if a.OIDC != nil {
```
nmiculinic, line 76: also check that one and only one field is set up (e.g. have a counter how many fields are setup, increase by one, and at the end verify that the counter is equal to 1).

aborilov: done

pkg/operator/internal/webhooks/kubecarrier_webhook.go
```diff
@@ -70,5 +71,40 @@ func (r *KubeCarrierWebhookHandler) validateCreate(kubeCarrier *operatorv1alpha1
 	if kubeCarrier.Name != constants.KubeCarrierDefaultName {
 		return fmt.Errorf("KubeCarrier object name should be 'kubecarrier', found: %s", kubeCarrier.Name)
 	}
+	auth := map[string]bool{}
```
jiachengxu, line 74: I think in go, we use map[string]struct{} as set in common cases

pkg/internal/resources/apiserver/apiserver.go
```diff
@@ -125,100 +124,105 @@ func Manifests(c Config) ([]unstructured.Unstructured, error) {
...
-						Name:  "HTPASSWD_SECRET_NAME",
-						Value: c.Spec.StaticUsers.HtpasswdSecret.Name,
-					},
-				)
-				supportedAuth = append(supportedAuth, "Htpasswd")
+	var supportedAuth []string
+	for _, auth := range c.Spec.Authentication {
+		if auth.StaticUsers != nil {
+			containers := deploymentPatch.Spec.Template.Spec.Containers
+			for i, container := range containers {
```
nmiculinic, line 131: wait...you have only a single container, you're explicitly controlling it since it's a local variable, and you're looping over it finding the manager container? Could you make this simpler by just doing something like:

```container := &deploymentPatch.Spec.Template.Spec.Containers[0]``` in code somewhere. (You want to get the pointer to it, for all changes be reflected to that container)

File: pkg/internal/resources/apiserver/apiserver.go [hunk 6]
```diff
@@ -125,100 +124,105 @@ func Manifests(c Config) ([]unstructured.Unstructured, error) {
...
-				)
-				containers[i].Env = append(containers[i].Env,
-					corev1.EnvVar{
-						Name:  "HTPASSWD_SECRET_NAME",
-						Value: c.Spec.StaticUsers.HtpasswdSecret.Name,
-					},
-				)
-				supportedAuth = append(supportedAuth, "Htpasswd")
+	var supportedAuth []string
+	for _, auth := range c.Spec.Authentication {
```
nmiculinic, line 128:
```suggestion
	var supportedAuth []string
	present := make(map[string]struct{})
	for i, auth := range c.Spec.Authentication {
		if auth.StaticUsers != nil {
			if _, ok := present["statik"]; ok {
				return nil, fmt.Errorf("statik config redefined")
			}
			present["statik"] = struct{}{}
			if err := addStatikConfig(auth.StaticUsers, &deploymentPatch.Spec.Template.Spec); err != nil {
				return nil, err
			}
			supportedAuth = append(supportedAuth, "Statik")
		}
		if len(supportedAuth) != (i + 1) {
			return nil, fmt.Errorf("only a single auth provider can be defined per statement. got %v", supportedAuth[i + 1:])
		}
		...
	}
```

perhaps extract each auth provider to its own method; it's easier to read and follow the main function. Also, you're missing the checks if only a single one is non-nil.

pkg/internal/resources/apiserver/apiserver.go
aborilov: >wait...you have only a single container, you're explicitly controlling it since it's a local variable, and you're looping over it finding the manager container? Could you make this simpler by just doing something like

I think it's your code, I just leave it as is :)
but yeah, I can simplify it

pkg/internal/resources/apiserver/apiserver.go
aborilov: >Also, you're missing the checks if only a single one is non-nil.

we have this check in the `kubecarrier_webhook`

aborilov:
> I've gone over the whole PR again. Please address the comment, I've left a few of my own. Afterward, I'll lgtm it; it's almost done except for a few nits/legibility here and there.

I think I've done all of them except defaulting webhook, as for now, we have only validation webhook? do you think it's a good idea to change it to defaulting webhook in this PR?

pkg/operator/internal/webhooks/kubecarrier_webhook.go
aborilov: it looks much nicer with `map[string]bool{}` then `map[string]strunct{}{}`.
also with `bool` as a value we can do `if auth['anonymous']` but with `struct` you will have to do `if _, ok := auth['anonymous']; ok{`

pkg/internal/resources/apiserver/apiserver.go
nmiculinic: defense in depth.

pkg/internal/resources/apiserver/apiserver.go
nmiculinic: > I think it's your code

What daemon possessed me? :P

pkg/internal/resources/apiserver/apiserver.go
aborilov: add we have `if-else` here, so only the first one will be added

nmiculinic:
I'm still worried about validation and error handling. Currently, only in the kubecarrier CRD, the API server validation is being checked. Not in the API Server CRD, which the user can create, nor in the `apiserver.Manifests` resource generation. If no/multiple auth options are specified per line in any one of those the failure shall be silent and ignored --> the most bite you in the ass type. 

What I'm suggesting is creating: `func validateAPIServerSpec(*operatorv1alpha1.APIServerSpec) error` and `defaultAPIServerSpec(*operatorv1alpha1.APIServerSpec) error`  helper functions which can and should be used in 3 places:

(maybe placing those functions close to the API definitions, or even make them member function for the operatorv1alpha1.APIServerSpec object? `.Validate() error` `.Default() error`) 

* Kubecarrier CRD webooks 
* API Server CRD webhooks  (this doesn't exists for now)
* `apiserver.Manifests` function (especially here, since it's the last point before silent failure) 

What do you think? 

Otherwise, this is a great PR, well written, and tested!

jiachengxu:
> I'm still worried about validation and error handling. Currently, only in the kubecarrier CRD, the API server validation is being checked. Not in the API Server CRD, which the user can create, nor in the `apiserver.Manifests` resource generation. If no/multiple auth options are specified per line in any one of those the failure shall be silent and ignored --> the most bite you in the ass type.

@nmiculinic The worrying is completely valid, and I agree that this may become an issue in the future, but for now I think it is fine. The reason are as following:
1. In the usecase that we offer KubeCarrier, users don't have permission to interact with operator API group, so this should be fine.
2. In the usercase that someone install KubeCarrier, and play around with it.  Mistakes in APIServerCRD and also in apiserver.Manifests will be kind of "ignored" , this is because KubeCarrier operator controller will reconcile the object with the default KubeCarrer config from CLI, and for that one we have a webhook to reject all mistakes.

aborilov:
I don't like to overcomplicate things, in our case users don't interact with APIserver configuration directly, so I don't think we need to add webhooks to apiserver operator. I will add defaulting webhook for KubeCarrier operator and I like the idea of `Default` and `Validate` methods for `APIServiceSpec`, will also do that. Later, if we decide to add webhooks for APIserver operator, we can easily use them

nmiculinic:
@aborilov , that's what I meant, adding APIserver webhooks is too much for this PR. Adding checks in `apiserver.Manifests` should be all I ask for. Those validate functions shall be called in two places -- kubecarrier webook and `apiserver.Manifests` and all is good!

aborilov:
> @aborilov , that's what I meant, adding APIserver webhooks is too much for this PR. Adding checks in `apiserver.Manifests` should be all I ask for. Those validate functions shall be called in two places -- kubecarrier webook and `apiserver.Manifests` and all is good!

yep, that's done

PR #516, opened 2020-07-23
Expose log level in KubeCarrier CRD

### LINKED ISSUE #485: Expose log level in KubeCarrier CRD
We want to be able to override the log level of KubeCarrier, so users can access debug or tracing level logs in order to troubleshoot issues. Users may also choose another log level to reduce the amount of logs KubeCarrier produces.

Component-level granularity of this setting is not strictly needed, but it would assist greatly in reducing the amount of extra logs that are generated when doing debug or tracing logs. And thus also prevents accidental congestion of the cluster log sink.

aborilov:
We want to be able to override the log level of KubeCarrier, so users can access debug or tracing level logs in order to troubleshoot issues. Users may also choose another log level to reduce the amount of logs KubeCarrier produces.

Fixes #485 

**Documentation**:

```release-note
Added `logLevel` field to KubeCarrier CRD
```


pkg/internal/resources/apiserver/apiserver.go
```diff
@@ -83,6 +84,7 @@ func Manifests(c Config) ([]unstructured.Unstructured, error) {
 								"--tls-cert-file=$(API_SERVER_TLS_CERT_FILE)",
 								"--tls-private-key-file=$(API_SERVER_TLS_PRIVATE_KEY_FILE)",
 								"--authentication-mode=$(AUTHENTICATION_MODE)",
+								"-v=$(VERBOSE)",
```
nmiculinic, line 87:
```suggestion
								"-v=$(LOG_LEVEL)",
```

same to other places as well

pkg/manager/internal/controllers/customresourcediscovery_controller.go
```diff
@@ -253,8 +254,10 @@ func (r *CustomResourceDiscoveryReconciler) reconcileCatapult(
 				Name: crDiscovery.Spec.ServiceCluster.Name,
 			},
 			WebhookStrategy: crDiscovery.Spec.WebhookStrategy,
+			LogLevel:        viper.GetInt("verbose"),
```
nmiculinic, line 257: This is too deep for the global variable lookup. It should be set in the controller's struct. 

And anyway, we discussed we're going to setup LogLevel in defaulting webhook if not otherwise stated.

pkg/apis/operator/v1alpha1/apiserver_types.go
```diff
@@ -62,6 +62,9 @@ type APIServerSpec struct {
 	// +optional
 	// Authentication configuration
 	Authentication Authentication `json:"authentication,omitempty"`
+	// LogLevel
+	// +optional
+	LogLevel int `json:"logLevel,omitempty"`
```
nmiculinic, line 67: This doesn't differentiate between set value to `0` and please default me. 

As we discussed 2 weeks ago, if the `LogLevel` is unset (that is `*int` is `null`) in the defaulting webhook we default to the operators `LogLevel`. Otherwise, we leave it as it is.

pkg/internal/resources/apiserver/apiserver.go
aborilov: done

pkg/manager/internal/controllers/customresourcediscovery_controller.go
aborilov: done

pkg/apis/operator/v1alpha1/apiserver_types.go
aborilov: done

go.mod
```diff
@@ -20,13 +20,14 @@ require (
 	github.com/grpc-ecosystem/go-grpc-middleware v1.2.0
 	github.com/grpc-ecosystem/go-grpc-prometheus v1.2.0
 	github.com/grpc-ecosystem/grpc-gateway v1.14.3
-	github.com/huandu/xstrings v1.3.0 // indirect
+	github.com/huandu/xstrings v1.3.2 // indirect
 	github.com/improbable-eng/grpc-web v0.12.0
 	github.com/jetstack/cert-manager v0.13.0
-	github.com/kubermatic/utils v0.0.0-20200707131849-09ebdec30295
```
nmiculinic, line 26: hmmm I don't see any tags in the utils repo. This should probably stay `v0.0.0-date-git_commit` format

pkg/internal/util/webhook/webhook.go
```diff
@@ -36,6 +37,14 @@ func IsDNS1123Label(s string) bool {
 	return dns1123LabelRegex.MatchString(s)
 }
 
+type LogLevelSetter interface {
+	SetLogLevel(int)
+}
+
+func SetDefaultLogLevel(spec LogLevelSetter) {
+	spec.SetLogLevel(viper.GetInt("verbose"))
```
nmiculinic, line 45: I'm heavily opposed fetching global variables this deep down the stack.

pkg/manager/internal/controllers/customresourcediscovery_controller.go
```diff
@@ -255,6 +255,7 @@ func (r *CustomResourceDiscoveryReconciler) reconcileCatapult(
 			WebhookStrategy: crDiscovery.Spec.WebhookStrategy,
 		},
 	}
+
```
nmiculinic, line 258: remove this line

pkg/operator/internal/webhooks/catapult_webhook.go
```diff
@@ -0,0 +1,74 @@
...
+	obj := &operatorv1alpha1.Catapult{}
+	if err := r.decoder.Decode(req, obj); err != nil {
+		return admission.Errored(http.StatusBadRequest, err)
+	}
+
+	switch req.Operation {
+	case adminv1beta1.Create:
+		r.Log.Info("creating catapult", "spec", obj.Spec)
+		if obj.Spec.LogLevel == nil {
+			webhook.SetDefaultLogLevel(webhook.LogLevelSetter(&obj.Spec))
```
nmiculinic, line 53: Why not fetch `kubecarrier` object directly and set its log value as default one? The object would already be in a cache, thus it's no network calls. Furthermore, this would negate the need for a deep global variable fetch.

pkg/internal/util/webhook/webhook.go
aborilov: it's public method, so it's not deep, it depends where are you call it and we are calling it in a webhooks, which is also not deep :)

pkg/cli/internal/cmd/setup/setup.go
```diff
@@ -129,15 +129,30 @@ func runE(conf *rest.Config, log logr.Logger, cmd *cobra.Command, flags *flags)
...
+		if yamlErr != nil {
+			return err
+		}
+	}
 	if err := spinner.AttachSpinnerTo(s, startTime, fmt.Sprintf("Create %q Namespace", ns.Name), createNamespace(ctx, c, ns)); err != nil {
 		return fmt.Errorf("creating KubeCarrier system namespace: %w", err)
 	}
 
-	if err := spinner.AttachSpinnerTo(s, startTime, "Deploy KubeCarrier Operator", reconcileOperator(ctx, log, c, ns)); err != nil {
+	if err := spinner.AttachSpinnerTo(s, startTime, "Deploy KubeCarrier Operator", reconcileOperator(ctx, log, c, ns, kubeCarrier.Spec.LogLevel)); err != nil {
```
jiachengxu, line 151: I don't really understand here, why do we set operator's logLevel from KubeCarrier CRD?
I think the operator logLevel should be passed as a flag in `kubecarrier setup` from user, and then we set up the log level of KubeCarrier CRD from that.
WDYT?

pkg/internal/resources/apiserver/apiserver.go
```diff
@@ -61,6 +62,10 @@ func Manifests(c Config) ([]unstructured.Unstructured, error) {
 		}
 	}
 	const AuthModeEnv = "AUTHENTICATION_MODE"
+	var logLevel int
+	if c.Spec.LogLevel != nil {
```
jiachengxu, line 66: I have checked the webhook implementation, webhook(defaulting of loglevel) will only be triggered in CREATE request, what if someone issues a UPDATE request with LogLevel nil? It will be set to "0". But I think it should still be the loglevel of Operator.

pkg/manager/internal/controllers/derivedcustomresource_controller.go
```diff
@@ -292,8 +292,9 @@ func (r *DerivedCustomResourceReconciler) Reconcile(req ctrl.Request) (ctrl.Resu
 		return result, nil
 	}
 
-	// leave `Paused` flag as is
+	// leave `Paused` flag and `LogLevel` as is
 	desiredElevator.Spec.Paused = currentElevator.Spec.Paused
+	desiredElevator.Spec.LogLevel = currentElevator.Spec.LogLevel
```
jiachengxu, line 297: This approach has a drawback, we need to find all the places that we update the Operator CRD objects, which is really hard to maintain. I think this can be solved by triggering webhook also during `UPDATE`, there is a
`req.OldObject` in the request, and you just check if the loglevel of the Oldobject is nil, if it is set, and just let it go., if it is not set, default it to operator's loglevel.

jiachengxu: This can also fix the issue here: https://github.com/kubermatic/kubecarrier/pull/516#discussion_r459950089

pkg/cli/internal/cmd/setup/setup.go
nmiculinic: My biggest use case for this feature is debugging certain operators, or deploy whole solution with extra logging.

I'm fine with either implementation, though I slightly prefer reusing the KubeCarrier CRD defined one.

pkg/internal/resources/apiserver/apiserver.go
nmiculinic: > But I think it should still be the loglevel of Operator.

I disagree, it should be default log level defined in the Kubecarrier object

pkg/manager/internal/controllers/derivedcustomresource_controller.go
nmiculinic: lgtm

pkg/internal/resources/apiserver/apiserver.go
jiachengxu: yeah, you are right, it should be default to Kubecarrier loglevel.

pkg/manager/internal/controllers/derivedcustomresource_controller.go
aborilov: I can also move `pause` flag logic also to the webhook `update` method, is it ok to do that in this PR?

pkg/internal/resources/apiserver/apiserver.go
aborilov: working on getting logLevel from `Kubecarrier` object

pkg/manager/internal/controllers/derivedcustomresource_controller.go
jiachengxu: For pause flag is that ok if we do it in a dedicated PR? This PR is getting larger and larger :)

pkg/cli/internal/cmd/setup/setup.go
jiachengxu: Ok, I am fine with it

pkg/internal/util/webhook/webhook.go
nmiculinic: It's public method within package, within package, within package. How big is the stack trace where this method gets called? 

What this does is the following:

* Tightly couples viper to the cmd implementation
* We need to register `verbose` flag
* We perform flag validation to `int` only when the webhook is called. (thus you could commit `GetString` and it would fail when the webhook is called, as opposed when program starts. Talk about failing fast.)
* It's difficult injecting test behavior for testing those functions under unit test (if we ever wrote some) since we depend on the global setting

Overall all the reasons why I dislike this approach. Have it as a field in the webhook struct (valid in other circumstances, not what I want from this PR), or fetch is from KubeCarrier object (cached) every time and set it to that value.

pkg/operator/internal/webhooks/catapult_webhook.go
aborilov: doen

pkg/manager/internal/controllers/customresourcediscovery_controller.go
aborilov: done

pkg/internal/util/webhook/webhook.go
aborilov: done

pkg/internal/resources/apiserver/apiserver.go
aborilov: done

pkg/internal/util/webhook/webhook.go
```diff
@@ -36,6 +42,24 @@ func IsDNS1123Label(s string) bool {
...
+}
+
+func SetDefaultLogLevel(ctx context.Context, c client.Client, spec LogLevelSetter) error {
+	kubeCarrier := &operatorv1alpha1.KubeCarrier{}
+	kubeCarrier.Name = constants.KubeCarrierDefaultName
+	objKey, err := client.ObjectKeyFromObject(kubeCarrier)
+	if err != nil {
+		return err
+	}
+	if err := c.Get(ctx, objKey, kubeCarrier); err != nil {
```
nmiculinic, line 56:
```suggestion
	kubeCarrier := &operatorv1alpha1.KubeCarrier{}
	if err := c.Get(ctx, types.NamespaceName{Name: constants.KubeCarrierDefaultName}, kubeCarrier); err != nil {
```

simplify this, no need for those extra lines

pkg/manager/internal/controllers/derivedcustomresource_controller.go
aborilov: > This approach has a drawback, we need to find all the places that we update the Operator CRD objects, which is really hard to maintain. I think this can be solved by triggering webhook also during `UPDATE`, there is a
> `req.OldObject` in the request, and you just compare the loglevel of the Oldobject with the Object, if it is same as the `OldObject`, and just let it go., if it is not set or not same, default it to OldObject loglevel.

in that way, you won't be able to change logLevel by your self

pkg/internal/util/webhook/webhook.go
aborilov: done

PR #525, opened 2020-07-28
Add common labels to service selector, to avoid service endpoint overlap

thetechnick:
When registering multiple CRDs in the same namespace the webhook selector will mach all elevator/catapult instances, do round robin load balancing and lead to this error:

```
Error from server (InternalError): error when creating "./master_cluster/kubecarrier/services/cert-svc1-tenant1.yaml": Internal error occurred: failed calling webhook "mcertificate.kubecarrier.io": Post https://certificates-svc-1-catapult-webhook-service.provider-1.svc:443/mutate-svc-1-provider-1-v1alpha2-certificate?timeout=30s: x509: certificate is valid for certificates-svc-2-catapult-webhook-service.provider-1.svc, certificates-svc-2-catapult-webhook-service.provider-1.svc.cluster.local, not certificates-svc-1-catapult-webhook-service.provider-1.svc
```

Fixes #

**Documentation**:

```release-note
NONE
```

### DIFF
--- a/pkg/internal/resources/catapult/catapult.go
+++ b/pkg/internal/resources/catapult/catapult.go
@@ -72,7 +72,7 @@ func Manifests(c Config) ([]unstructured.Unstructured, error) {
 		CommonLabels: map[string]string{
 			constants.NameLabel:      "catapult",
 			constants.InstanceLabel:  c.Name,
-			constants.ManagedbyLabel: constants.ManagedbyKubeCarrierOperator,
+			constants.ManagedByLabel: constants.ManagedByKubeCarrierOperator,
 		},
 		Resources: []string{"../default"},
 		PatchesStrategicMerge: []types.PatchStrategicMerge{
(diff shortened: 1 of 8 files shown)


pkg/internal/resources/catapult/catapult.go
```diff
@@ -294,9 +299,6 @@ func Manifests(c Config) ([]unstructured.Unstructured, error) {
 		if labels == nil {
 			labels = map[string]string{}
 		}
-		labels[constants.NameLabel] = "catapult"
-		labels[constants.InstanceLabel] = c.Name
-		labels[constants.ManagedbyLabel] = constants.ManagedbyKubeCarrierOperator
```
jiachengxu, line 299: These lines cannot be removed, otherwise, some objects will not have common labels, because not all objects are created from `MkLayer`
You can check the changes of `catapult.golden.yaml`, and the common labels are removed from a `ClusterRole` which we created after the `MkLayer`

thetechnick: Ok re-added that code

PR #530, opened 2020-07-30
Add generator for example objects from a JSON Schema

thetechnick:
This PR adds code that generates examples of CRDs registered in KubeCarrier.

e.g.

This Schema:
```yaml
type: object
properties:
  apiVersion:
    type: string
  kind:
    type: string
  spec:
    type: object
    properties:
      prop1:
        type: string
      prop2:
        type: string
        default: test123
      stringArray:
        type: array
        items:
          type: string
      roles:
        type: array
        items:
          properties:
            name:
              type: string
```

Will result in this example object (when also adding GVK):

```yaml
apiVersion: test.kubecarrier.io/v1alpha1
kind: Test
metadata:
  name: test
spec:
  prop1: ""
  prop2: test123
  roles:
  - name: ""
  stringArray: []
```

This generated example object provides a quick overview of the API (at least easier and quicker than going through the json schema by hand) and can be used to prefill the create dialog in the KubeCarrier UI.

Fixes #

This function could be implemented in 3 different places:
- Within the `DerivedCustomResource` controller (adding to status) and copy the information to `Offering` objects in the `Catalog` controller
- In the API Server (don't really like this/needs recalculation every request)
- Not in KubeCarrier -> Clients need to do this

**Documentation**:

```release-note
```

### DIFF
--- a/pkg/manager/internal/controllers/util.go
+++ b/pkg/manager/internal/controllers/util.go
@@ -17,9 +17,14 @@ limitations under the License.
 package controllers
 
 import (
+	"encoding/json"
+	"strings"
+
 	apiextensionsv1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/runtime/schema"
 )
 
 const (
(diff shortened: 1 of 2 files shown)


jiachengxu:
I would like to do it within the DerivedCustomResource controller because I think our APIServer is a lightweight interface that should not have complicated logic.

thetechnick:
I actually have not a big opinion about where to put this functionality, except that I concur with @jiachengxu that the API server is probably the worst place.

@nmiculinic Putting it into the UI is certainly an option, although I think it's kinda useful for everyone and the overhead is neglectable. Also I would prefer to keep such stuff out of the UI, until we commit to a longer standing code base for it...

The reason why I think this is very useful is, because parsing the JSONSpec in your head is taking way longer than looking at the example and prone to indentation errors and typos.

nmiculinic:
> DerivedCustomResource

As you already know, I want to kill it; thus wouldn't agree to put it there.

> The reason why I think this is very useful is, because parsing the JSONSpec in your head is taking way longer than looking at the example and prone to indentation errors and typos.

well...yes in away. Though let's first tackle bigger issues over nice usability sugars?

PR #531, opened 2020-07-30
Deprecate KindOverride fields and use internal api group.

jiachengxu:

Fixes #

**Documentation**:

```release-note
Deprecate KindOverride fields and use internal api group.
```


pkg/manager/internal/controllers/customresourcediscovery_controller_test.go
```diff
@@ -135,7 +135,7 @@ func TestCustomResourceDiscoveryReconciler(t *testing.T) {
 
 	internalCRD := &apiextensionsv1.CustomResourceDefinition{}
 	require.NoError(t, r.Client.Get(ctx, types.NamespacedName{
-		Name: strings.Join([]string{"redis", serviceClusterName, crDiscovery.Namespace}, "."),
+		Name: strings.Join([]string{"redis", "internal", serviceClusterName, crDiscovery.Namespace}, "."),
```
aborilov, line 138: why not using constant here?

test/integration/apiserver.go
```diff
@@ -380,14 +378,14 @@ func instanceService(ctx context.Context, conn *grpc.ClientConn, tenantAccount *
 		offering := &catalogv1alpha1.Offering{
 			ObjectMeta: metav1.ObjectMeta{
 				Namespace: tenantAccount.Status.Namespace.Name,
-				Name:      strings.Join([]string{"dbs", serviceCluster.Name, providerAccount.Name}, "."),
+				Name:      strings.Join([]string{"dbs", "external", serviceCluster.Name, providerAccount.Name}, "."),
```
aborilov, line 381: same here

pkg/manager/internal/controllers/customresourcediscovery_controller_test.go
jiachengxu: fixed here.

test/integration/apiserver.go
jiachengxu: We cannot use constants pkg in test, since constants are under `internal` package.

aborilov: that is sad

jiachengxu: hmm, if needed, we can move the whole constants package out of internal, what do you think?

aborilov: it's not a big deal, I'm ok to merge it as is, but yeah, I think constants should not be in the internals

jiachengxu: OK, let's do it in a dedicated PR and move the whole constants pkg out of internal

aborilov: 👍

PR #536, opened 2020-08-03
Increase make generate performance on Mac by using delegated mount mode

jiachengxu:
The current implementations of mounts on Linux provide a consistent view of a host directory tree inside a container: reads and writes performed either on the host or in the container are immediately reflected in the other environment, and file system events (inotify, FSEvents) are consistently propagated in both directions.
On Linux, these guarantees carry no overhead, since the underlying VFS is shared directly between host and container. However, on macOS (and other non-Linux platforms) there are significant overheads to guaranteeing perfect consistency.
This PR using `delegated` (the container’s view is authoritative (permit delays before updates on the container appear in the host)) to increase the performance of `make generate` on MacOS.
Ref: https://docs.docker.com/docker-for-mac/osxfs-caching/
Fixes #

**Documentation**:

```release-note
NONE
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -118,7 +118,7 @@ generate-grpc:
 ifdef CI
 	@hack/proto-codegen.sh
 else
-	@docker run --rm -e CI=true -w /src -v $(PWD):/src \
+	@docker run --rm -e CI=true -w /src -v $(PWD):/src:delegated \
 		--user "$(id -u):$(id -g)" \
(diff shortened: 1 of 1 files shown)


nmiculinic:
Does this have any drawbacks to linux?

jiachengxu:
> Does this have any drawbacks to linux?

As far as I know, nope

PR #537, opened 2020-08-03
KEP-001 Remove derive custrom resource

nmiculinic:
```release-note
NONE
```

### DIFF
--- a/docs/proposals/kep-001-derive-support-removal.md
+++ b/docs/proposals/kep-001-derive-support-removal.md
@@ -1,0 +1,73 @@
+# [KEP-001] Derived CR support removal 
+
+## Background
+
+Currently, the workflow for propagating custrom resource from the service cluster to the managment cluster's tenant, and back works like this:
+
+* custom resource is 2-way-propagated from service cluster to the management cluster via the ferry. Its group is translated to the managment clusters internal group. (i.e. `internal.svc-1.provider-1`)
+* Such exposed custom resource passes through derivation step, one this proposal aims removing. The derivation step serves two purposes:
+    * filtering which custom resource fields are exposed
+    * exposing it via offering to the tenants
+    
+The original aim for derivation is lightweight service offering templating and field filtering by the service provider to its consumers, that is kubecarrier tenants. 
+
+For example a database offering could original be like this on the service cluster:
+
+```yaml
+apiGroup: db.io
+kind: Redis
+metadata:
+    name: example
+    namespace: tenant-1sfsd1
+spec:
+    instanceSize: large
+status:
+    url: "redis://example.db.io"
+    phase: "Ready"
+    conditions: [...]
+```
+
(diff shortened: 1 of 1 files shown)


thetechnick:
**Reason for DerivedCustomResource**
The main reason for the DerivedCR is, that we want to provide central service management for service operators (providers) and service consumers (tenants).
In my experience, working for multiple service providers, this is provided within an internal and an external API surface and attached tooling.

This way support agents, operations, etc. can access additional information about a service or trigger actions not exposed on the public API.

Additional information includes:
- Detailed status information
- Links to internal monitoring systems
- Deployment details (real replica number, etc)

Actions/Properties
- Setting the version of an instance
- Setting the replicas directly
- Forcing restarts
- Forcing relocation

**Issues**
Multi cluster awareness in the operator and network setup is neither improved nor worsened by the derive process.

Otherwise these points remain:
- maintenance burden for KubeCarrier development (it's complicated/a fair bit of code involved)
- maintenance burden for service operators (the configuration needs to be created and maintained)

Maintenance burden for service operators is not really a lot IMHO, as service operators already need to add metadata about the service (descriptions/icons/etc)

Maintenance burden for KubeCarrier development is somewhat huge, although we already have the code and tests running and ongoing support of the feature, should be less hassle.

Overall I am not convinced that we should just drop this feature, but I am very much open to discuss other ideas to solve the underlying issue.

nmiculinic:
You're raising valid points, one came to my mind as I was writing this proposal. (It was still important writing it though).

> Maintenance burden for service operators is not really a lot IMHO, as service operators already need to add metadata about the service (descriptions/icons/etc)

All these extra steps should be optional for easier getting started/playing around with. IMHO derive is nice to have, albeit tad incomplete for now. 

> Maintenance burden for KubeCarrier development is somewhat huge, although we already have the code and tests running and ongoing support of the feature, should be less hassle.

Let's just remind ourselves the product hasn't been battle-tested, and we'll see how many unknowns unknows pop up on this code path. The most bug-free code is no code at all. We'll see how often is this feature used.

nmiculinic:
> I think most latencies should be cross-cluster propagation? Which we will have anyways, propogation from internal CRD to external CRD happens in a single cluster, should not matter that much?

have you measured it? Because the difference in propagation lag is one IP package travel time, which for clusters in the same datacenter is negligible (~1ms, maybe ~10ms). Bigger propagation delay is c/p all the state and ensuring the state view is actually consistent. #542  touches on it. 

> Well, if DerivedCustomResource is dropped, there should be some other object, for example, CustomResourceDiscovery will need to do filtering/derivation work, which doesn't reduce the maintenance burden.

this proposal is against doing filtering/derivation work in the first place, not renaming it under a different banner.

aborilov:
I think we can make atleast derive configuration optional, exposing everything by default, it will make it easier to start to work with kubecarrier

PR #538, opened 2020-08-03
KEP-002 Bulward integration

nmiculinic:
```release-note
NONE
```

### DIFF
--- a/docs/proposals/kep-002-bulward.md
+++ b/docs/proposals/kep-002-bulward.md
@@ -6,7 +6,7 @@ Currently, we have `Account` concept in kubecarrier, which encompasses single us
 
 ## Migration plan
 
-Current `Account` concept closely matches bulward's `Project`. It's a sandbox where single focused responsibility hosts its resources. Thus, all references to the account through the kubecarrier solution should be renamed to account. In the following code excerpt you can see similarities and differences.
+Current `Account` concept closely matches bulward's `Project`. It's a sandbox where single focused responsibility hosts its resources. Thus, all references to the account through the kubecarrier solution should be renamed to project. In the following code excerpt you can see similarities and differences.
 
 ```yaml
 apiVersion: catalog.kubecarrier.io/v1alpha1
--- a/docs/proposals/kep-002-bulward.md
+++ b/docs/proposals/kep-002-bulward.md
@@ -51,7 +51,7 @@ spec:
 
 ### Differences
 
-* `Account` is namespace scoped, while the `Project` is namespaces scopes, within the `Organization` namespace. `Organization` itself is a cluster scoped object.
+* `Account` is cluster scoped, while the `Project` is namespaces scopes, within the `Organization` namespace. `Organization` itself is a cluster scoped object.
 * `Project` object currently has no metadata concept, though it should be added in further bulward version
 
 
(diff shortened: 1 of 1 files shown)


docs/proposals/kep-002-bulward.md
```diff
@@ -0,0 +1,102 @@
...
+```
+
+### Similarities
+
+* both have ownership concept
+* both define a namespace where the owners & memeber operation take place
+
+### Differences
+
+* `Account` is namespace scoped, while the `Project` is namespaces scopes, within the `Organization` namespace. `Organization` itself is a cluster scoped object.
```
jiachengxu, line 54: `Account` is a cluster-scoped object.

thetechnick:
Overall ok, although I think you should work on wording. There are a lot of "shall"s in there.

Two bigger topics:
- We cannot use labels/annotations on the `Project` object to assign Tenant/Provider roles to it, as users should not be able to give themselves the provider or tenant role in KubeCarrier.

- I would like to think about deployment of bulward separately. In the end, bulward should by able to be deployed on it's own and plug into both KubeCarrier and Kubermatic.

jiachengxu:
I have created another proposal of bulward integration, #543 , please let me know what do you think

docs/proposals/kep-002-bulward.md
nmiculinic: fixed

docs/proposals/kep-002-bulward.md
```diff
@@ -0,0 +1,107 @@
+# [KEP-002] Bulward integration
+
+## Background
+
+Currently, we have `Account` concept in kubecarrier, which encompasses single user. Due to bigger interop with the Kubermatic ecosystem, as well as further requirements on the user management we built `bulward`. The account is permeated through the kubecarrier project, as it's central user concept. The API server's proto file specify the account, the accounts come into two possible roles (Provider and Tenant), and adequate RBAC permissions
+
+## Migration plan
+
+Current `Account` concept closely matches bulward's `Project`. It's a sandbox where single focused responsibility hosts its resources. Thus, all references to the account through the kubecarrier solution should be renamed to account. In the following code excerpt you can see similarities and differences.
```
jiachengxu, line 9:
```suggestion
Current `Account` concept closely matches bulward's `Project`. It's a sandbox where single focused responsibility hosts its resources. Thus, all references to the account through the kubecarrier solution should be renamed to project. In the following code excerpt you can see similarities and differences.
```

docs/proposals/kep-002-bulward.md
```diff
@@ -0,0 +1,107 @@
...
+    name: team-a-member
+    apiGroup: rbac.authorization.k8s.io
+```
+
+```yaml
+apiVersion: apiserver.bulward.io/v1alpha1
+kind: Project
+metadata:
+  name: project-example
+  namespace: organization-a
```
jiachengxu, line 36: Do we require the Organization to be created before creating Project? Because Project objects are namespaced-scoped, we should think about if we force to create Organization first

nmiculinic: that's a bulward thing...I'd say yes. Even if you have a single root organization.

nmiculinic: btw, we now have defined arbitrary hierarchy two levels deep without recursion (projects are part of the organization). We currently don't support recursion (i.e. projects cannot contain multiple sub-project) nor deeper hierarchies.

Issue #316, opened 2020-03-23
[Feature] CustromResourceDiscovery/DerivedCustromResource explicit propagation


nmiculinic:
**Is your feature request related to a problem? Please describe.**
Currently, the `CustromResourceDiscover`/`DerivedCustromResource` propagation policy between base resource and the derived/copied resource is a bit ad-hoc in a way. Its propagation policy is implicit (has it originated as in an original base resource object or as a derived/copy object, which fields are sync and in which direction across the boundaries, etc). Furthermore, this is currently an async process, where it maybe shouldn't be. 

**Describe the solution you'd like**
For this issue, I propose the following solution. Both the `CustomResourceDiscovery` and `DerivedCustromResource` should have a `.spec.policy` field with two possible options: 

* `Forward`  (the default one)
* `Backward`

With the following semantic:

For the `Forward` propagation policy from the derived/copy object the following things are propagated into the base resource object:
* all non-status & non-metadata fields
* The following metadata fields
    * name
    * namespace (In DerivedCustromResource case, otherwise not)
    * labels
    * annotations

And from the base resource the following attributes are synced back to the derived/copy object:
* status 
* finalizers (to be determined whether by aggregate finalizers or copy finalizers by value)

The base object shouldn't be explicitly modified by other controllers than the kubecarrier except for the status field. The derived/copy object should be explicitly modified only for the propagated fields as per the above specification. The object must begin its life as a derived/copy object and the kubecarrier's controller shall create its base object pair. There aren't any adoption controller or that complexity handling these operations. The base obj shouldn't be explicitly deleted, only by the kubecarrier controllers. The derived/copy object could be deleted explicitly, which would trigger the underlying reconciliation. The derived/copy object should be present until the underlying base object is deleted and cleared up from the cluster; that is a foregroundDeletion operation. 

As for the `Backward`, the primary mechanism would be an immutable read-only view of the base object. That means from the base object the following attributes are propagated to the derived/copy object:
* all non-metadata fields
* The following metadata fields
    * name
    * namespace (In DerivedCustromResource case, otherwise not)
    * labels?
    * annotations?
    * finalizers? 

[...]

PR #542, opened 2020-08-04
KEP-003 service cluster resource API Aggregation

nmiculinic:
```release-note
NONE
```

See also https://github.com/kubermatic/kubecarrier/issues/316

### DIFF
--- a/docs/proposals/kep-003-service-resouce-aggregation.md
+++ b/docs/proposals/kep-003-service-resouce-aggregation.md
@@ -2,9 +2,9 @@
 
 ## Background
 
-Currently, we're creating ferry deployment per each `CustromResourceDiscovery` object. The ferry component transports custom resources from the service cluster to the management cluster, and vice versa, asynchronously. For example the `db.example.io` apiGroup in the service cluster `svc-1` becomes `db.internal.svc-1.provider-1` apiGroup in the management cluster.
+Currently, we're creating catapult deployment per each `CustromResourceDiscovery` object. The catapult component transports custom resources from the service cluster to the management cluster, and vice versa, asynchronously. For example the `db.example.io` apiGroup in the service cluster `svc-1` becomes `db.internal.svc-1.provider-1` apiGroup in the management cluster.
 
-Furthermore, ferry component handles the webhooks for internal apiGroup. If there is `CustromResourceDiscoverySet` there's a new ferry deployment per each service cluster offering the same service.
+Furthermore, catapult component handles the webhooks for internal apiGroup. If there is `CustromResourceDiscoverySet` there's a new catapult deployment per each service cluster offering the same service.
 
 * See also:
 
@@ -37,7 +37,7 @@ The issues with this approach are multi-fold:
 
 ## Proposed solution
 
-Instead of creating CRDs for each `CustomResourceDiscovery`, and deploying ferry per `CustomResourceDiscovery`, we deploy a single `neoferry` per provider. This `neoferry` is API Exstension server smartly proxying each API group to the service cluster, as well aggregating multiple service cluster under same apiGroup. The benefits are as follows:
+Instead of creating CRDs for each `CustomResourceDiscovery`, and deploying catapult per `CustomResourceDiscovery`, we deploy a single `neocatapult` per provider. This `neocatapult` is API Exstension server smartly proxying each API group to the service cluster, as well aggregating multiple service cluster under same apiGroup. The benefits are as follows:
 
 * sync propagation --> there's no version divergence, we only provide a view in the management cluster with custom RBAC sprinkled on top.
 * webhook complexity --> Under the hood original webhooks are ran on the service cluster, thus we're staying as close to the true north as possible.
--- a/docs/proposals/kep-003-service-resouce-aggregation.md
+++ b/docs/proposals/kep-003-service-resouce-aggregation.md
@@ -1,0 +1,57 @@
+# [KEP-003] Service resource aggregation
+
+## Background
+
+Currently, we're creating ferry deployment per each `CustromResourceDiscovery` object. The ferry component transports custom resources from the service cluster to the management cluster, and vice versa, asynchronously. For example the `db.example.io` apiGroup in the service cluster `svc-1` becomes `db.internal.svc-1.provider-1` apiGroup in the management cluster.
+
+Furthermore, ferry component handles the webhooks for internal apiGroup. If there is `CustromResourceDiscoverySet` there's a new ferry deployment per each service cluster offering the same service.
+
+* See also:
+
+https://github.com/kubermatic/kubecarrier/issues/316
+
+## Issues
+
+The issues with this approach are multi-fold:
+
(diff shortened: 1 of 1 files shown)


thetechnick:
- We don't have a ferry component any more, it was replaced by elevator and catapult.
- When referring to other concepts/issues, please explain how they interact (prerequisite, related bug, ...)

## 2/3-way merges

We are not reinventing a git-like multi branch process.

Kubernetes Controllers have a strict hierarchy, where higher level controllers use the APIs of lower level controllers. When KubeCarrier controls an instance, the management cluster is the source of truth and overrides all other changes.

If properties (like replicas in your example) are changed by another controller, they need to be hidden from the higher level controllers, as other controllers cannot "posses" the field. In the future more complex scenarios could maybe be handled via field ownership...

## Sync Propagation

When doing sync calls to the underlying cluster, we are running into the classical distributing computing issue of adding N*Critical dependencies that degrade the main systems uptime.

In praxis, this will result in network errors or cluster maintenance disabling the user facing API. Create/update/delete/show APIs will break as soon as the specific backend cluster is unreachable, while LIST request require that all clusters are reachable all the time.

Even small network interruptions to some clusters (which are quite frequent in big systems, especially across cloud provider/data center boundaries), would degrade the KubeCarrier API to an unusable point.

nmiculinic:
> We don't have a ferry component any more, it was replaced by elevator and catapult.

my mistake, I was thinking of catapult when writing it. Fixed.

> 2/3-way merges
> Kubernetes Controllers have a strict hierarchy, where higher level controllers use the APIs of lower level controllers. When KubeCarrier controls an instance, the management cluster is the source of truth and overrides all other changes.

Well..yes and no. What bout the adoption controller? It's my pet peeve which doesn't fit nicely in this world view. As currently implemented once the new adopted object is created in the service cluster (e.g. etcd Snapshot) it gets "owned" by the management cluster despite originating in the service cluster, and a service cluster operator is its original creator.

In case the service cluster operator wants to update this adopted object, especially changing its spec, it would get overwritten by the management cluster. 

> When doing sync calls to the underlying cluster, we are running into the classical distributing computing issue of adding N*Critical dependencies that degrade the main systems uptime.

True. Though unless network partition happens between management cluster and service cluster things should be fineish ( service cluster has HA kube-apiserver as well as HA etcd installation. Probably.)

In case the service cluster kube-apiserver is unreachable, should we really not error out on that? Not let our users know there are issue, but mask those issues like snake hides its legs? (I know we're letting them know on the `ServiceCluster` object, but this is more direct. You gain immediate information things are broken, retry again later) 

If we're still pursuing eventual consistency, what you think about earlier proposal https://github.com/kubermatic/kubecarrier/issues/316 regarding differentiating objects where the management cluster is source of truth, versus where the service cluster is a source of truth?

docs/proposals/kep-003-service-resouce-aggregation.md
```diff
@@ -0,0 +1,57 @@
...
+https://github.com/kubermatic/kubecarrier/issues/316
+
+## Issues
+
+The issues with this approach are multi-fold:
+
+* propagation is asynchronously, whereas management clusters serves as a "cache" to service cluster's ground truth. In more complicated setups the management cluster's resource version and service cluster resource version could diverge in unreconciable way. To give a simple, albeit toy example consider the following:
+
+    * There's deployment on the service cluster. This resource is propagated to the management cluster. At the start the deployment replicas is 1.
+       * HPA controller on the service cluster increases the replica count to 2
```
jiachengxu, line 20: What is HPA controller?

jiachengxu: Who increases the replica count? The service provider modify the object in the service cluster directly or?

docs/proposals/kep-003-service-resouce-aggregation.md
```diff
@@ -0,0 +1,57 @@
...
+       * User on the management cluster changes the deployment image
+
+   Now we have to two diverging versions what the deployment should be, and how should it look like. Somehow we have to perform 2-way, or 3-way merge between those objects adding additional complexity.
+
+   Note: In the common case the service cluster operators would only change the `.status` subresource, while the management cluster would handle the `.spec` field.
```
jiachengxu, line 25: Yeah, agree, and I think in general, operators/controller should not modify `spec` fields? and Webhook may default some fields.

docs/proposals/kep-003-service-resouce-aggregation.md
```diff
@@ -0,0 +1,57 @@
...
+
+* webhook complexity
+
+    We're running semi-complex code for object updating and creation webhooks. We rely on `dry-run` support from the service cluster. Service cluster resource might not support some critical `dry-run` operation, thus our webhooks won't behave as if they are being run in the service cluster in the first place.
```
jiachengxu, line 29: What do you mean that "resource might not support some critical `dry-run` operation"?
But yeah, you are right, the KubeCarrier dynamic webhooks require some extra conifgurations of service cluster webhook, for example, the `SideEffects` should be configured to `NoneOnDryrun`, otherwise it will reject request with dry-run flag.
ref: https://github.com/kubermatic/kubecarrier/pull/396

docs/proposals/kep-003-service-resouce-aggregation.md
```diff
@@ -0,0 +1,57 @@
...
+* sysadmin operator automation
+    * Where `CustomResourceDiscoverySet` is used for exposing same service from multiple service cluster, there's no unified resource on the management cluster for automation. There's `db.internal.svc-1.provider-1`, `db.internal.svc-2.provider-1`, ... api groups. The sysadmin wants something simple, `db.internal.provider-1` apiGroup smartly aggregating each service cluster resource into an unified apiGroup.
+
+* resource consumption
+    * multiple deployments per `CustomResourceDiscovery` start stacking up, increasing the KubeCarrier solution heaviness and limiting efficient resource allocation. Lean is green.
+
+
+## Proposed solution
+
+Instead of creating CRDs for each `CustomResourceDiscovery`, and deploying catapult per `CustomResourceDiscovery`, we deploy a single `neocatapult` per provider. This `neocatapult` is API Exstension server smartly proxying each API group to the service cluster, as well aggregating multiple service cluster under same apiGroup. The benefits are as follows:
```
jiachengxu, line 40: How do we dynamically register new apis to the `neocatapult` for provider?

docs/proposals/kep-003-service-resouce-aggregation.md
nmiculinic: HorizontalPodAutoscaler  

https://kubernetes.io/docs/tasks/run-application/horizontal-pod-autoscale/

docs/proposals/kep-003-service-resouce-aggregation.md
nmiculinic: `APIServices` resource? 

This is unknown unknown (read I haven't tried it), but I think I can add additional APIs during runtime to the API extension server

docs/proposals/kep-003-service-resouce-aggregation.md
nmiculinic: Exactly that, there's some webhook without the `SideEffect: NoneOnDryRun` which would run upon creation, but it's actually mislabeled `dry-run`.

PR #543, opened 2020-08-06
Proposal Bulward Integration

jiachengxu:

Fixes #

**Documentation**:

```release-note
NONE
```

### DIFF
--- a/docs/proposals/bulward-integration.md
+++ b/docs/proposals/bulward-integration.md
@@ -128,7 +128,9 @@ There are two possible solutions:
 - We can have dedicated `Project` for both `couchdb` and `redisdb` (extra propagation tooling needed).
 - These objects can just stay in the `Organization` namespace, and owner set different permissions for different teams by
 creating `OrganizationRole`.
+
 This pretty much depends on Organization Owners.
+
 Here is an example if owners wants to create dedicated `Project`:
 ```yaml
 apiVersion: apiserver.bulward.io/v1alpha1
--- a/docs/proposals/bulward-integration.md
+++ b/docs/proposals/bulward-integration.md
@@ -67,6 +67,12 @@ spec:
 There are few reasons that we don't replace the `Account` with `Project`:
 - KubeCarrier cannot take fully advantage of Organization features in Bulward. For big enterprises, `Organzation` concept
 is needed for them to manage permissions of different team, different user has different permission for different project.
+- `Organization` is the top level in Bulward, and organization owners have freedom to manage Projects as they like, for example,
+moving objects around to `Project` namespaces to ease the management and resource isolation. While `Project` is not splittable,
+which means there is no sub-project in `Project`.
+- Services that available for the whole `Organization` can be shared by different `Project`s. Catalog related objects can
+be put in a central place (`Organization` namespace), can be shared by `Project`s.
+so if we map `Account` to `Project`,
 - `Account` is cluster-scoped object, while `Project` is namespace-scoped object, integration will introduce lot of changes.
 
 So this proposal proposes the following changes:
@@ -118,7 +124,12 @@ spec:
     name: redisdb-operator
     apiGroup: rbac.authorization.k8s.io
 ```
-And then we can have dedicated `Project` for both `couchdb` and `redisdb`:
+There are two possible solutions:
+- We can have dedicated `Project` for both `couchdb` and `redisdb` (extra propagation tooling needed).
+- These objects can just stay in the `Organization` namespace, and owner set different permissions for different teams by
+creating `OrganizationRole`.
+This pretty much depends on Organization Owners.
+Here is an example if owners wants to create dedicated `Project`:
 ```yaml
 apiVersion: apiserver.bulward.io/v1alpha1
 kind: Project
@@ -153,8 +164,7 @@ we can use `ProjectRoleTemplate` and `ProjectRole` to set up permission for a (g
 
 Overall:
 - `Account` should be mapped to `Organzation` in a way, not `Project`.
-- `Organization` can be `Provider` and `Tenant` at the same time, but `Project` under
-`Organization` should be defined as Provider project or Tenant project.
+- `Organization` can be `Provider` and `Tenant` at the same time.
 
 ## Implementation
 ### Deployment
@@ -185,11 +195,11 @@ by `OrganizationRoleTemplate`, KubeCarrier should have permission to update/crea
 permission degradation will be handled by Bulward. Also, organization owner or rbac-admin can set different perssmions
 for different project by using `ProjectRole`.
 #### Account Replacement
-Let's go back to `company-a` example above, `couchdb` and `redisdb` are two projects under the same
+Let's go back to `company-a` example above, `couchdb` and `redisdb` are two services under the same
 `Organization`, and `couchdb` is project to consume couchdb services from other KubeCarrier user, and `redisdb` is a project
 to provide `redisdb` services from `company-a` to other users.
-In our controllers, we replace the namespace of `Account` with the namespace of `Project`, which means:
-For `couchdb`, is a tenant project, so member of this project should be able to work with:
+In our controllers, we replace the namespace of `Account` with the namespace of `Organization`, which means:
+For `couchdb`, is a tenant project, so member of this team should be able to work with:
 - offerings
 - providers
 - regions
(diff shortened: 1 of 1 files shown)


thetechnick:
Some questions from me:

- How do do we configure that a Project is providing/consuming services?
- When we offer something to an organization, should the catalog with all it's objects just be present in the organization namespace? -> No need to copy it into every project namespace.
- Does KubeCarrier really need to know which Project is marked as a Tenant/Provider, or can we let the Org-Owner setup permissions completely free and deal with whatever namespace stuff(service instances/service cluster/catalogentrysets) appear in?

jiachengxu:
> * How do do we configure that a Project is providing/consuming services?

I think this should be configured by Organization Owner/Project admin who has permission to work with Project, they can add labels to `Project` objects.
> * When we offer something to an organization, should the catalog with all it's objects just be present in the organization namespace? -> No need to copy it into every project namespace.

When we offer something(for example redisdb) to an organization, the Organization owner should create a Project (consum-redisdb, for example) to house all redisdb related stuff. And then Organization owener/project owner can add members to this project, so they can consume the redisdb service, does that make sense to you?

> * Does KubeCarrier really need to know which Project is marked as a Tenant/Provider, or can we let the Org-Owner setup permissions completely free and deal with whatever namespace stuff(service instances/service cluster/catalogentrysets) appear in?

Yeah, Org owner can do that, but I think if KubeCarrier knows whihc Project is Tenant/Provider, it can automate that for Owner :) WDYT?

docs/proposals/bulward-integration.md
```diff
@@ -0,0 +1,206 @@
...
+`Project` is namespace-scoped, and similar to `Organization`, every `Project` will have a namespace.
+
+### Account <-> Organization or Project?
+There are few reasons that we don't replace the `Account` with `Project`:
+- KubeCarrier cannot take fully advantage of Organization features in Bulward. For big enterprises, `Organzation` concept
+is needed for them to manage permissions of different team, different user has different permission for different project.
+- `Account` is cluster-scoped object, while `Project` is namespace-scoped object, integration will introduce lot of changes.
+
+So this proposal proposes the following changes:
+We use `Organziation` to represent the `Account` in KubeCarrier. For better explanation, here I will provide an example of
```
nmiculinic, line 73: This is kinda a moot point. I'd argue most `Accounts` would be either pure tenants or pure providers. There would be a minority being both a tenant and a provider, which IMHO would be providers with service dependencies.

docs/proposals/bulward-integration.md
```diff
@@ -0,0 +1,206 @@
...
+Every bulward release can contain a manifest yaml file, just like cert-manager, we can install it in the similar way:
+```bash
+kubectl apply -f https://github.com/kubermatic/bulward/releases/download/${BULWARD_VERSION}/bulward.yaml
+kubectl wait --for=condition=available deployment/bulward-controller-manager -n bulward-system --timeout=120s
+kubectl wait --for=condition=available deployment/bulward-apiserver -n bulward-system --timeout=120s
+```
+### API Server
+#### Organization Endpoint
+We have an endpoint for `Account`, we can reuse it for `Organization`, also, in Bulward, only `Organization` object is visible
+for organization member (this is implemented in Bulward api extension server), we can also drop the filtering implementation in currently account endpoint.
```
nmiculinic, line 170: Yes and no. 

Correct me if I'm wrong, in our current API server implementation we use the same REST config for our k8s client, without any user impersonation? Thus when we query `apiserver.bulward.io` we would have same visibility as API server service account. 

Thus we either need to query `storage.bulward.io` and do API server filtering ( a duplicate work, and bad approach ). Or get the user REST config impersonation working per API server backend request.

docs/proposals/bulward-integration.md
```diff
@@ -0,0 +1,206 @@
...
+#### Project Endpoint
+Project Endpoint (Get/List/Watch/Update/Delete) should be implemented, all permission filtering are offloaded to Bulward api
+extension server, so this will also be a thin layer.
```
nmiculinic, line 173: plus we have to inject the user impersionation in the API server as well.

docs/proposals/bulward-integration.md
jiachengxu: > Correct me if I'm wrong, in our current API server implementation we use the same REST config for our k8s client, without any user impersonation?

Yes, that' right. 
In our `Account` endpoint, we have a filtering layer now, we use `FieldIndexer` to fetch the `Account` objects, which has user name in the `Account.Spec.Subjects`.

When we integrate KubeCarrier with Bulward, as you mentioned, I would like to use an impersonation client, so we can cut the current filtering layer in API server, and offload that to Bulward completely.

docs/proposals/bulward-integration.md
jiachengxu: yeah, that's right, will update the proposal

docs/proposals/bulward-integration.md
nmiculinic: we could also use the same impersonation to replace or complement our authZ layer as well.

docs/proposals/bulward-integration.md
jiachengxu: done

aborilov:
> When we offer something(for example redisdb) to an organization, the Organization owner should create a Project (consum-redisdb, for example) to house all redisdb related stuff. And then Organization owener/project owner can add members to this project, so they can consume the redisdb service, does that make sense to you?

I don't think it's a real-world example, usually, organizations want to consume several services in their project, for example for some web applications it can be services like redis+nginx+postgresql.

jiachengxu:
> I don't think it's a real-world example, usually, organizations want to consume several services in their project, for example for some web applications it can be services like redis+nginx+postgresql.

This is just an example :)
In this case, organization owner create a project to house redis+nginx+postgresql, and add members to this project. 
Does that make sense to you?

aborilov:
> > I don't think it's a real-world example, usually, organizations want to consume several services in their project, for example for some web applications it can be services like redis+nginx+postgresql.
> 
> This is just an example :)
> In this case, organization owner create a project to house redis+nginx+postgresql, and add members to this project.
> Does that make sense to you?

yeah, and the question was, `should the catalog with all it's objects just be present in the organization namespace? `. So do we need to have all catalog entries in the Project namespace? or users can get it from organization namespace?

jiachengxu:
> should the catalog with all it's objects just be present in the organization namespace?
I think this depends on Organization Ower/Admin, how they want to organization projects/objects stuff. If they think there are too many objects in the Organziation namespace, and it's not easy for management, they can move them into sub projects.

docs/proposals/bulward-integration.md
```diff
@@ -67,6 +67,12 @@ spec:
 There are few reasons that we don't replace the `Account` with `Project`:
 - KubeCarrier cannot take fully advantage of Organization features in Bulward. For big enterprises, `Organzation` concept
 is needed for them to manage permissions of different team, different user has different permission for different project.
+- `Organization` is the top level in Bulward, and organization owners have freedom to manage Projects as they like, for example,
+moving objects around to `Project` namespaces to ease the management and resource isolation. While `Project` is not splittable,
+which means there is no sub-project in `Project`.
+- Services that available for the whole `Organization` can be shared by different `Project`s. Catalog related objects can
```
nmiculinic, line 73: This seems a bit messy. Thus you're allowing catalogs to live both in the organization as well as in the projects? What's the benefit in that over replicating it in each project. I'm viewing project as a closed sandbox where all project related stuff are contained.

jiachengxu: Nope, as I mentioned in the proposal, this is not a mandatory thing, and it depends on the organization owners.
1. Organization owners can keep all catalog stuff in the Organization namespace, and create `OrganizationRole` to mange members.
2. If Organization owners want to have more resource isolation, they can create Project, and move objects into Project namespace.

However, if we map Project to Account, we kinda lose the possibility of 1, i.e., Catalog cannot be shared in a Organization, and every Project will have a copy of catalog objects in namespace.

aborilov: how `Organizations` catalogs combined with `Projects` catalogs? I mean, if the project doesn't have any catalogs in its namespace, should we check organizations namespace for catalogs or this means that project don't have any catalogs?

rastislavs:
Closing for now, the proposal will be re-considered later.

PR #547, opened 2020-08-11
Add a experimental conversion step for NonStructuralSchema

### LINKED ISSUE #546: [Bug] CRD propagation with version v1beta1
**Describe the bug**
Currently, KubeCarrier only supports propagation of CRD with version `v1` from service cluster to management cluster. This is because `v1` CRD has some different validations than `v1beta1`, and in KubeCarrier controllers we work with `v1` CRD directly. It can still work with the `v1beta1` CRDs which can pass the validation of `v1` when do conversion.

thetechnick:

Fixes #546

**Documentation**:

```release-note
```


thetechnick:
This will not work easily, too many conversion issues.
To properly handle this, we need to let the kubeapiserver do the conversion, which will be quite complicated to do...

